PR #15, opened 2022-07-07
add cloudTrail S3 Script

williamputraintan:
- Create query and retrieve query result from cloudTrail lake per given principalId, and eventDataStore (additional filter could be timestamp and key)
- Query S3 object size to see if egress bytes match with events
- Print results in table as example below
	Screen Shot 2022-07-07 at 4 22 57 pm

Currently CloudTrail lake is created manually. Could do it in a cdk, but I would assume would be done in the elsa deployment setup?

We could identify if the data transferred is via cli or pre-signed url (via the `userAgent`), but I think if we create the accessKeyId unique per sharing, then I don't think it would be relevant to distinguish those events as most likely the sharing would either be pre-signed or the s3-cli


andrewpatto:
Excellent work - thoughts on whether we can get this being used in the short term for tracking AG data sharing? @ohofmann  is the output of this enough to be useful for you?
Do you have a nodejs runtime installed (or could install it?)
We can probably bundle it up into a util you can run..

application/util/cloudtrail_s3/src/main.ts
```diff
@@ -0,0 +1,235 @@
...
+  GetQueryResultsCommand,
+} from "@aws-sdk/client-cloudtrail";
+
+/**
+ * AWS Client
+ */
+const CONFIG = {
+  credentials: fromIni({ profile: process.env.AWS_PROFILE }),
+  region: "ap-southeast-2",
+};
```
andrewpatto, line 15: I tend to use a completely blank config `{ }` and assume it picks up stuff from the env - do we have any consensus on how we should be doing this @victorskl ? I would think insisting on a fromIni() means that this same code _won't_ be able to be run in a lambda for instance?

ohofmann:
Spotted this yesterday - yes, very helpful. What is the input to this currently? A unique tag/access key embedded in the URLs of a single data access request or something else?

application/util/cloudtrail_s3/src/main.ts
victorskl: Yup; we could/should defer it to SDK at all, as much as possible. 

Will; could you try like this; to see whether it works.

```
const credentials = new AWS.SharedIniFileCredentials();
```

i.e. SDK will do the look up in some well-known order. Please see [note here](https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/setting-credentials-node.html).

victorskl: Also. Please ditto [this example](https://github.com/umccr/data-portal-apis/blob/dev/docs/examples/portal_api_sig4.js) Sig4 signing code, if any.

andrewpatto:
Because we are using SSO and not static IAM - the access key changes each (AWS) session

andrewpatto:
I mean we could make a map of access keys and the corresponding sharing request ID in this repo - the access keys themselves aren't secrets. 
Long term the idea is that Elsa retains this access key -> release map in its db

application/util/cloudtrail_s3/src/main.ts
victorskl: (Ah) Just realise that I am still pointing to ini cred file. 😁 Please check [AWS.CredentialProviderChain](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/CredentialProviderChain.html) there, actually.

Good stuff, btw!

williamputraintan: > const credentials = new AWS.SharedIniFileCredentials();

Yes, this for the SDK v2. I used the [SDK v3 ](https://docs.aws.amazon.com/en_us/sdk-for-javascript/v3/developer-guide/setting-credentials-node.html)

williamputraintan: > I tend to use a completely blank config { }

I tried this before, but somehow giving me an expired token error while doing say `aws s3 ls` does work

```
/Users/wpintan/UMCCR/GitHub/elsa-data/application/util/cloudtrail_s3/node_modules/@aws-sdk/client-cloudtrail/dist-cjs/protocols/Aws_json1_1.js:1808
            response = new CloudTrailServiceException_1.CloudTrailServiceException({
                       ^

ExpiredTokenException: The security token included in the request is expired
    at deserializeAws_json1_1StartQueryCommandError (/Users/wpintan/UMCCR/GitHub/elsa-data/application/util/cloudtrail_s3/node_modules/@aws-sdk/client-cloudtrail/dist-cjs/protocols/Aws_json1_1.js:1808:24)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async /Users/wpintan/UMCCR/GitHub/elsa-data/application/util/cloudtrail_s3/node_modules/@aws-sdk/middleware-serde/dist-cjs/deserializerMiddleware.js:7:24
    at async /Users/wpintan/UMCCR/GitHub/elsa-data/application/util/cloudtrail_s3/node_modules/@aws-sdk/middleware-signing/dist-cjs/middleware.js:11:20
    at async StandardRetryStrategy.retry (/Users/wpintan/UMCCR/GitHub/elsa-data/application/util/cloudtrail_s3/node_modules/@aws-sdk/middleware-retry/dist-cjs/StandardRetryStrategy.js:51:46)
    at async /Users/wpintan/UMCCR/GitHub/elsa-data/application/util/cloudtrail_s3/node_modules/@aws-sdk/middleware-logger/dist-cjs/loggerMiddleware.js:6:22 {
  '$fault': 'client',
  '$metadata': {
    httpStatusCode: 400,
    requestId: 'b5c4018c-3058-4f82-a714-f01e862e8145',
    extendedRequestId: undefined,
    cfId: undefined,
    attempts: 1,
    totalRetryDelay: 0
  },
  __type: 'ExpiredTokenException'
}
```

victorskl: Ah. Right! v3. I have yet to reckon with. Sorry.

victorskl:
Just to note. IIRC. We normally use a dedicated IAM user for PreSigning purpose. So they won't be changed quite often as SSO would on every login. 

However, there could be potentially that the IAM user's credential will be rotated/revoked, yes. So, there is still needing a release map store some where for historical record keeping.

andrewpatto:
Is the dedicated IAM user to that we can get long(er) lived pre-signed keys?

williamputraintan:
> Is the dedicated IAM user to that we can get long(er) lived pre-signed keys?

Yes, IAM user give a maximum full 7 days access. Ref: [aws-blog](https://aws.amazon.com/premiumsupport/knowledge-center/presigned-url-s3-bucket-expiration/#:~:text=To%20create%20a%20presigned%20URL%20that's%20valid%20up%20to%20seven,in%20AWS%20Signature%20Version%204.)

Other IAM may still have `expires=...` within 7 days in the url query parameters, but I believe the access is still limited to their maximum IAM session.

victorskl:
Also for the Audit trail. Same scenario here. So that we could track S3 object access pattern. It was a bit of ad-hoc activity in the past. But, I reckon, this will become more routine trail now.

andrewpatto:
Hmm.. I was hoping to use the access key to distinguish between different releases of data.. I mean if the rate of data release is minimal we can have a bunch of IAM users or something I guess.

I know that extra parameters can be signed into a pre signed url.. I wonder whether it is possible for us to add a custom param (Elsa=ABCD) and get that into the cloud trail log?

victorskl:
> add a custom param (Elsa=ABCD)

Might be possible. Something custom Sig4 signing process on canonical path, I think. Not really sure though. We could explore. 😁

victorskl:
> a bunch of IAM users or something I guess.

Yah. May be we can do like per IAM user per cohort/flagship release (or something). One high level common view. 

> use the (varying) access key to distinguish between different releases of data

Or, another thinking is -- since one IAM user can have multiple access keys; per release per access key still doable that way, under some "common" IAM user. As long as we do not revoke it (either remove or make it inactive), etc.

andrewpatto:
> Or, another thinking is -- since one IAM user can have multiple access keys (not sure upper bound though); per release per access key still doable that way

Unfortunately the upper bound is 2 I think (from memory)

victorskl:
> upper bound is 2

Ouch!

williamputraintan:
> I know that extra parameters can be signed into a pre signed url.. I wonder whether it is possible for us to add a custom param (Elsa=ABCD) and get that into the cloud trail log?

Possible, one of the cloudtrail logs is `requestParameters` which might appear there. I'll explore that if that is the case.

williamputraintan:
Seems to be correct! CloudTrail does record all parameter including those custom one. Added `hello=world` param in the presignedUrl

```
{X-Amz-Date=20220708T063518Z, bucketName=examp-temp-dev, X-Amz-Algorithm=AWS4-HMAC-SHA256, X-Amz-SignedHeaders=host, Host=examp-temp-dev.s3.ap-southeast-2.amazonaws.com, X-Amz-Content-Sha256=UNSIGNED-PAYLOAD, X-Amz-Expires=900, hello=world, key=example/test-head-1.csv}
```

williamputraintan:
Just wondering if we decided to use copy command (aws s3 cp), no custom parameter will be present there.

andrewpatto:
Nice - does that mean that the hello=world actually participated in the signing algorithm? (if you change to hello=earth in the pre-signed url text does it fail?)

williamputraintan:
Yes, it will fail

andrewpatto:
> Just wondering if we decided to use copy command (aws s3 cp), no custom parameter will be present there.

Yes I think we'd do AWS<->AWS direct copies slightly differently.. possibly I'd make a per-release S3 access point. And track access that way.

application/util/cloudtrail_s3/src/main.ts
williamputraintan: Found the issue, we could leave the config blank `{ }`. There were conflict in `AWS_PROFILE` and it prioritize `shared credentials file` instead of the valid creds (that logged in via `sso-login`) which in this case uses the expired creds-file .

andrewpatto (approved):

andrewpatto:
Same here - other than perhaps moving it out of the application folder to a top-level util/ folder - happy to just get this merged and we can refactor as we go

PR #22, opened 2022-07-21
Edgedb RDS integration

williamputraintan:
- Infrastructure init
- Deployment elsa edgedb server with hostname of `db.elsa.dev.umccr.org`
- Allow edgedb to consume TLS certificate pre-defined before the stack


application/infrastructure/bin/elsa.ts
```diff
@@ -0,0 +1,23 @@
...
+import { ElsaStack } from "../stack/elsa-stack";
+
+const context = {
+  namespace: "elsa",
+};
+
+const app = new cdk.App({ context: context });
+
+const stackTag = {
+  stack: "elsa",
```
andrewpatto, line 13: I think to match our UMCCR Convention for tags - this might need to be "Stack" (actually I have no idea if aws tags are case sensitive??)

andrewpatto:
Also as a general thing - I think I'd like 'infrastructure' to be up one level in the directories.. so I want "application" to be Elsa only (i.e. the software absent any deployment stuff)

And then at the top of the repo - something like
infrastructure/aws
infrastructure/google
etc

application/infrastructure/bin/elsa.ts
williamputraintan: Yes it is! Changing it

andrewpatto:
Happy to just merge this in direct - and we'll review/refactor it as we go forward with the project..

andrewpatto (approved):

PR #49, opened 2022-09-09
Render DUO consent popups as something more than JSON

### LINKED ISSUE #31: Render DUO consent popups as something more than JSON
Each node (case / patient / specimen / dataset) can have an accompanying Consent declaration. At the moment this is an array of DUO code JSON structures.

In the "cases" box - when someone hovers over a node - we allow that consent to be fetched and displayed in a popup. See below.

Screen Shot 2022-08-24 at 9 21 57 am

It would be nice to tidy this popup up - so that the consent statement has icons / text etc - and no JSON displays.

Skies the limit - for instance for a "countries" restriction - we could display a row of little flags! Or just the names of the countries!

Here are some icons that were donated to the DUO project that might also be useful.

https://github.com/EBISPOT/DUO/tree/master/doc/figs/pictograms/duo_png

DoxasticFox:
Resolves #31. Demo:
Untitled


application/frontend/src/pages/releases/detail/cases-box/consent-popup.tsx
```diff
@@ -52,10 +79,46 @@ export const ConsentPopup: React.FC<Props> = ({ releaseId, nodeId }) => {
...
       onOpen={onOpenHandler}
     >
-      <div className="flex flex-row divide-x divide-blue-500 bg-white text-xs border rounded drop-shadow-lg">
-        {duos.map((duo) => (
-          <pre>{JSON.stringify(duo, null, 1)}</pre>
-        ))}
+      <div className="p-2 space-y-4 bg-white text-sm border rounded drop-shadow-lg">
+        {duos.map(function(duo: DuoLimitationCodedType) {
+          const diseaseCode = (duo as any)?.diseaseCode;
+          const diseaseSystem = (duo as any)?.diseaseSystem;
```
DoxasticFox, line 86: I'm not thrilled about the type assertions, but I couldn't see a type-safe way to do it which wasn't very verbose. Suggestions?

andrewpatto: Yeah agreed - I did the type definitions of DUO structures (which I think is valuable as a schema) but have never been happy about how to use them in practice at the react end.. lets keep thinking about this

andrewpatto:
Looks good. I think this is totally useable as the base - I'm wondering if we could add in a small amount of ontology resolution - in particular the DUO codes are such as small codeset - we can literally just have them as a static text lookup in the frontend. (i.e. turn "DUO:00007" to "Disease Specific Research (DUO:00007)")

And also thinking about asynchronous disease lookup using Ontoserver (same as the way it is done in the Disease Chooser component - turning MONDO:0037940 into the actual disease) - though that ontology lookup code is a mess currently so happy if we put that on the back burner as a different ticket.

andrewpatto:
All lookups can be in a 'best effort' mode - so if the code doesn't seem to exist or can't be resolved in time - then falling back to the exact display you have currently is ok.

DoxasticFox:
@andrewpatto do you know how I can fix this?

```
~/elsa-data % ( cd application/frontend ; npm run build:dev )
> elsa-data-frontend@0.0.0 build:dev
> react-scripts build

Creating an optimized production build...
Failed to compile.

Module parse failed: Unexpected token (26:7)
File was processed with these loaders:
 * ./node_modules/source-map-loader/dist/cjs.js
You may need an additional loader to handle the result of these loaders.
| export * from "./schemas-testing";
| 
> export type ReleaseApplicationCodedType = Static<
|   typeof ReleaseApplicationCodedSchema
| >;
```

Despite `react-scripts` not saying which file it's referring to, grepping matches `application/common/elsa-types/index.ts`. That doesn't seem to be the real problem though; I didn't modify that file and deleting code around line 26 just produces the same errors about adjacent code.

The actual problem seems to be this snippet in `application/frontend/src/pages/releases/detail/cases-box/consent-popup.tsx`:

```typescript
          const description
            = isKnownDuoCode(diseaseCode)
            ? duoCodeToDescription[diseaseCode]
            : null;
```

Replacing it with `const description = "";` allows the code to compile. Neither `npm run build:dev` nor my IDE produces errors about that code however, so I don't know why there's a problem with it.

andrewpatto:
I think its possibly that the elsa-types project is not really set up correctly for it to make declarations of actual content (like dictionaries) .. possibly.. (though they seem to work in elsa-constants .. ??)

TBH - the Typebox stuff I did for the DUO schemas is overly complex - I was kind of hoping to get it to do a whole bunch of automated type checking but it is way overkill. So possibly I've led you astray with that whole thing.

So I just moved the DUO -> descrption dictionary into the frontend itself and simplified the types (string -> string). That seems to have fixed the compile issue.

Along the way I also updated the typescript compilers in various package.json in an attempt to fix it..

DoxasticFox:
Thanks. So am I good to merge this as-is and pick up the Ontoserver stuff as a separate ticket?

andrewpatto:
Yes sure (I didn't actually run the code - just got it to compile) - but yes if the code adds the DUO rendering then happy for you to make a new ontoserver related one and close this.

andrewpatto (approved):

PR #58, opened 2022-09-16
Import DB from AG S3 (issue#43)

### LINKED ISSUE #43: Import the cardiac flagship index into Elsa as a dataset
Need the ability to point Elsa at a bucket/dynamo corresponding to the cardiac flagship and have Elsa safely create/update the dataset records for it.

Due to cardiac being an ongoing flagship (data still arriving) it would be great if this was a totally idempotent restartable operation - that picked up any changes between the current representation in Elsa and the current state of the buckets.

There is no problem if this operation takes a while (minutes).

If data is misisng/invalid (missing one of a pair) I think it is acceptable for this to display a warning and skip - it doesn't need to be Elsas job to fix the data. It can then be fixed in the bucket and this import run again to 'correct/insert' the fixed entries.

williamputraintan:
Attempt on importing AG dataset from S3 objects (issue #43).

Note [UPDATE 19/9]:
- The pairing of artifacts is based on s3Url. For FASTQ forward and reverse file, it will do filename manipulation to remove `_R1` or `_R2` from filename to determine if they belong together. This could be a problem if the filename itself will contain these character part of the filename.
-  If a pair of files are on a separated submission/prefix, it will skipped as consequent pair file will not be found. Could change this implementation if need be where pairing files are located on a different prefix. 
  (Initially pairing implementation is only by filename, but some filename are the same as part of reanalysis but located on a different prefixes)


andrewpatto:
I think the R1 R2 issue is ok being solved this way - but is a good one to note..

application/backend/src/business/db/lab-queries.ts
```diff
@@ -0,0 +1,147 @@
...
+  forwardFile: File,
+  reverseFile: File
+) {
+  return e.insert(e.lab.ArtifactFastqPair, {
+    forwardFile: insertFile(forwardFile),
+    reverseFile: insertFile(reverseFile),
+  });
+}
+
+export function insertArtifactBamQuery(bamFile: File, baiFile: File) {
```
andrewpatto, line 37: Not for actioning - just for thinking - I wonder whether we need some naming conventions to distinguish these from the later Edgedb query consts..

Maybe these are snippets or fragments? insertArtifactBamFragment..

Just musing..

application/backend/src/business/db/lab-queries.ts
```diff
@@ -0,0 +1,147 @@
...
+  return e.insert(e.lab.ArtifactCram, {
+    cramFile: insertFile(cramFile),
+    craiFile: insertFile(craiFile),
+  });
+}
+
+/**
+ * SELECT function queries
+ */
+export type ArtifactStudyIdAndFileIdByDatasetCaseIdQueryType = {
```
andrewpatto: There is some typescript magic that will auto generate this type btw.. again not for actioning as I can't remember how - but I've seen examples in the edgedb typescript docs

application/backend/src/business/services/aws-helper.ts
```diff
@@ -0,0 +1,75 @@
...
+
+          s3ObjectList.push({
+            key: key ?? "",
+            eTag: objContent.ETag ?? "",
+            size: objContent.Size ?? 0,
+          });
+        }
+      }
+    } catch (e) {
+      console.error(e);
```
andrewpatto: we probably want this to not be caught - unless we are going to attempt a Retry pattern.. if any part of the import fails I imagine it should abort the whole thing so this can just be thrown (I think I have the Fastify routing set up correctly such that an exception thrown will convert to a 500 HTTP response with details)

application/backend/tests/service-tests/ag.common.ts
```diff
@@ -0,0 +1,228 @@
+import e, { storage } from "../../dbschema/edgeql-js";
+
+/**
+ * Will declare some mock test data here
```
andrewpatto, line 4: I like the mock test setup here.. is good you've gone to the effort on this..

andrewpatto (approved):

application/backend/src/test-data/insert-test-data.ts
```diff
@@ -26,8 +26,7 @@ export async function insertTestData(settings: ElsaSettings) {
   await insert10G();
   await insert10F();
   await insert10C();
-  // cardiac is disabled because we now have a unique filename restriction - need to sort out
-  // await insertCARDIAC();
+  await insertCARDIAC();
```
andrewpatto, line 29: we can have a discussion when I'm back - not having any concrete thoughts on this yet - but just noting that we might end up with 'dev data' that is stuff that gets auto inserted for our dev purposes.. and 'test data' that is used for unit tests. At the moment these are kind of shared.. but could increasingly start to be split (as in the test data will tend to be more focussed on helping out with one specific unit test set and should be pretty unchanging - whereas the dev data we might change week to week to give demos etc).. so philiosophically could start to diverge

williamputraintan: Yup, I enabled it back as I found the problem that there were duplicated filename and the restriction is not allowing this.

application/backend/src/business/services/aws-helper.ts
williamputraintan: Yup, make sense!

williamputraintan:
> I think the R1 R2 issue is ok being solved this way - but is a good one to note..

An additional note on pairing files, some file have the same filename on a different prefixes. Will now consider on file s3Url instead of just filename.

PR #60, opened 2022-09-23
Render MONDO codes

### LINKED ISSUE #55: Expand MONDO codes in consent pop-ups
This ticket comes from Andrew Patterson's comment [here](https://github.com/umccr/elsa-data/pull/49#issuecomment-1243092077):

>And also thinking about asynchronous disease lookup using Ontoserver (same as the way it is done in the Disease Chooser component - turning MONDO:0037940 into the actual disease) - though that ontology lookup code is a mess currently so happy if we put that on the back burner as a different ticket.

DoxasticFox:
Resolves https://github.com/umccr/elsa-data/issues/55.

asdf

### DIFF
--- a/application/frontend/src/pages/releases/detail/cases-box/consent-popup.tsx
+++ b/application/frontend/src/pages/releases/detail/cases-box/consent-popup.tsx
@@ -12,7 +12,9 @@ import {
 } from "@fortawesome/free-solid-svg-icons";
 
 import {
+  CodingType,
   DuoLimitationCodedType,
+  DuoModifierType,
   ReleaseCaseType,
   ReleasePatientType,
 } from "@umccr/elsa-types";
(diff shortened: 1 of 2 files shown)


application/frontend/src/pages/releases/detail/cases-box/consent-popup.tsx
```diff
@@ -52,6 +66,38 @@ const Flags: React.FC<FlagsProps> = ({ regions }) => {
...
+  if (mondoSystem === undefined) {
+    return undefined;
+  }
+  if (mondoCode === undefined) {
+    return undefined;
+  }
+
+  const oldCodes: CodingType[] = [{system: mondoSystem, code: mondoCode}];
+  const newCodes = await doBatchLookup(
+    "https://onto.prod.umccr.org/fhir",
```
andrewpatto, line 90: We will have to 'config' ise this URL at some point - was my laziness of not doing it on the first Batch lookup code - but that can definitely be for another ticket

andrewpatto (approved):

application/frontend/src/pages/releases/detail/cases-box/consent-popup.tsx
DoxasticFox: I raised a ticket: https://github.com/umccr/elsa-data/issues/61

PR #63, opened 2022-10-12
Tidying of AWS deploy and edgedb settings

andrewpatto:
An attempt to add a serverless aurora (works) - but is to expensive for dev
Split out classes for some of the resources now that CDK constructs are reflected in CloudFormation panel
Lots of cert/secrets work - mostly unsuccesful
Added a dev mode that makes various artifacts public


andrewpatto:
Not particularly needing deep review - is basically the existing code but rejigged (split the load balanacer from the ecs because CDK only allows playing with some settings with these split)

williamputraintan:
Yup, LGTM!

williamputraintan (approved):

PR #64, opened 2022-10-12
Docker image for AWS deploy

andrewpatto:
There were some weird issues with audit log service and DI
But lots of general fixes and tinkering


andrewpatto:
Still basically WIP - so there are some docker bits that need tightening (it currently works both as ts-node code AND as esbuild in case we need to pivot back and forth)
But I wanted a build that basically worked for a AWS deployment demo next week

williamputraintan (approved):

PR #65, opened 2022-10-17
Issue#7 PresignedUrl File Update

### LINKED ISSUE #7: Create an initial download zip for a researcher that contains S3 links as per current AG process
A researcher should be able to login and hit a button that gives them a zip file with pre-signed S3 links as per the current python script (should we actually call the python script? (I think I would prefer not - but it depends on how much work has gone in to it)

williamputraintan:
- [FE] Minor CSS update
- [FE] Added checkbox to have a dynamic header selection for presignedUrl file
- [BE] `/api/releases/:rid/aws-s3-presigned` will now need a list of header for the presignedUrl file
- [BE] Include `x-releaseId=abcd-efgh` as a custom params in presignedUrl

### DIFF
--- a/application/backend/src/business/services/aws-presigned-urls-service.ts
+++ b/application/backend/src/business/services/aws-presigned-urls-service.ts
@@ -1,7 +1,6 @@
 import { AuthenticatedUser } from "../authenticated-user";
-import { GetObjectCommand, S3Client } from "@aws-sdk/client-s3";
+import { S3Client } from "@aws-sdk/client-s3";
 import * as edgedb from "edgedb";
-import { getSignedUrl } from "@aws-sdk/s3-request-presigner";
 import { inject, injectable, singleton } from "tsyringe";
 import { UsersService } from "./users-service";
 import { AwsBaseService } from "./aws-base-service";
@@ -11,6 +10,11 @@ import archiver, { ArchiverOptions } from "archiver";
 import { stringify } from "csv-stringify";
 import streamConsumers from "node:stream/consumers";
 import { ReleaseService } from "./release-service";
+import { HttpRequest } from "@aws-sdk/protocol-http";
+import { S3RequestPresigner } from "@aws-sdk/s3-request-presigner";
+import { parseUrl } from "@aws-sdk/url-parser";
+import { Hash } from "@aws-sdk/hash-node";
+import { formatUrl } from "@aws-sdk/util-format-url";
 
 @injectable()
 @singleton()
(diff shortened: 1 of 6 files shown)


application/frontend/src/pages/releases/detail/aws-s3-presigned-form.tsx
```diff
@@ -21,13 +31,32 @@ export const AwsS3PresignedForm: React.FC<Props> = ({
 }) => {
   const displayPassword = releaseData.downloadPassword ?? "(ask PI)";
 
+  const [isHeaderSelectionOpen, setIsHeaderSelectionOpen] =
+    useState<boolean>(false);
+
+  const [haederSelected, setHaederSelected] = useState<string[]>([
```
andrewpatto: typo in name

application/frontend/src/pages/releases/detail/aws-s3-presigned-form.tsx
```diff
@@ -1,13 +1,23 @@
...
+  "patientId",
+  "s3Signed",
+  "caseId",
+  "fileType",
+  "specimenId",
+  "size",
+  "s3Url",
+  "s3Bucket",
+  "s3Key",
+];
```
andrewpatto, line 20: rather than being repeated - can these in some way go into the shared typescript projects? (between backend and frontend).. I mean the types like FileRecordHeaderType might differ between front end and backend - but the allowable string constants should be common

application/frontend/src/pages/releases/detail/aws-s3-presigned-form.tsx
```diff
@@ -21,13 +31,32 @@ export const AwsS3PresignedForm: React.FC<Props> = ({
...
+  };
   return (
     <>
       {/* we use a POST form action here (rather than a onSubmit handler) because
             form POSTS can be converted natively into a browser file save dialog
              i.e. if the POST returned a Content-Disposition header */}
       <form
-        action={`/api/releases/${releaseId}/aws-s3-presigned`}
+        // onSubmit={handleSubmit}
+        action={`http://localhost:3000/api/releases/${releaseId}/aws-s3-presigned`}
```
andrewpatto: ?

application/frontend/src/pages/releases/detail/aws-s3-presigned-form.tsx
williamputraintan: Fixing it!

application/frontend/src/pages/releases/detail/aws-s3-presigned-form.tsx
williamputraintan: Sorry, wasn't supposed to be committed in the first place. Was trying to run FE on a different host to get the benefit of react build watcher on the development server, and this point to the BE localhost.

application/frontend/src/pages/releases/detail/aws-s3-presigned-form.tsx
williamputraintan: Yup, so I managed to try this by exporting the existing variable in the common folder ([#AwsFileRecordKey](https://github.com/umccr/elsa-data/blob/61d84d6e291d0dcbb061f319fe4a3013a21b989d/application/common/elsa-types/schemas.ts#L29) or [#FileRecordHeader](https://github.com/umccr/elsa-data/blob/61d84d6e291d0dcbb061f319fe4a3013a21b989d/application/common/elsa-types/schemas.ts#L42)) and import it on the frontend.

It does not seems to be working with loader when building the FE code.
```
> elsa-data-frontend@0.0.0 build:dev
> react-scripts build

Creating an optimized production build...
Failed to compile.

Module parse failed: Unexpected token (26:7)
File was processed with these loaders:
 * ./node_modules/source-map-loader/dist/cjs.js
You may need an additional loader to handle the result of these loaders.
```

Perhaps I am missing something?

application/backend/src/business/services/aws-presigned-urls-service.ts
```diff
@@ -49,19 +53,31 @@ export class AwsPresignedUrlsService extends AwsBaseService {
...
+
+      const presigner = new S3RequestPresigner({
+        credentials: awsCredentials,
+        region: awsRegion,
+        sha256: Hash.bind(null, "sha256"),
       });
+      const url = formatUrl(
+        await presigner.presign(new HttpRequest(s3ObjectUrl), {
+          expiresIn: 60 * 60 * 24 * 7, // 7 days
+        })
```
andrewpatto, line 78: just as a heads up we are going to need to do something about the permissions here - because even if we ask for 7 days, the role that Elsa Data has is a limited lifetime role (max 24 hrs)..

andrewpatto: not for this ticket though

application/frontend/src/pages/releases/detail/aws-s3-presigned-form.tsx
andrewpatto: No, I think I've seen a similar message.. will investigate as it does work for some constants.. fine for this ticket though.

andrewpatto (approved):

application/backend/src/business/services/aws-presigned-urls-service.ts
williamputraintan: Yes, would probably need a designated User credentials for Elsa ?

PR #70, opened 2022-10-20
Some minor fixes to get the deployed dev working correctly

andrewpatto:
The esbuild process breaks DI so I am pivoting to just running the server via ts-node (same as actual local dev)


application/frontend/src/pages/releases/detail/logs-box/logs-box.tsx
```diff
@@ -46,6 +46,10 @@ export const LogsBox: React.FC<Props> = ({ releaseId, pageSize }) => {
   const baseColumnClasses = "py-2 font-small text-gray-400 whitespace-nowrap";
 
   const createRows = (data: AuditEntryType[]) => {
+    // whilst it shouldn't happen in real life - in dev we can occasionally have
+    // releases with no audit entries - so we need to handle that safely
+    if (!data) return [];
+
```
andrewpatto, line 52: This was causing an issue in some of my dev scenarios

andrewpatto:
This code has actually been deployed to the current dev deployed instance

application/frontend/src/pages/releases/detail/logs-box/logs-box.tsx
williamputraintan: Yes, I suppose this is the property map does not exist (or something ... )?

williamputraintan (approved):

application/frontend/src/pages/releases/detail/logs-box/logs-box.tsx
mmalenic: Wouldn't the map operation below handle an empty array as well? Or does this check if the array is undefined?

williamputraintan: I think this checks for `undefined`, `.map` should handle OK for empty array.

mmalenic: Ahh yeah, makes sense.

mmalenic (approved):

PR #72, opened 2022-10-21
Create a cacheing *frontend* ontology lookup service

### LINKED ISSUE #50: Create a cacheing *frontend* ontology lookup service
We are using proper coding systems for things like gene pickers and disease codes. This will use a service like "ontoserver" direct from the front end code.

There is an example of its use from the Disease Chooser react component (where it looks up mondo diseases).

I cobbled together a horrible async/cache version in a hurry to get a demo working - but it needs to be done better.

I am thinking a proper cache (rather than a dictionary as it currently is)

https://github.com/isaacs/node-lru-cache

(or other??)

And the ability to pre-populate with static sets (ala country codes and DUO codes - which are codesets that we probably won't ever have in ontoserver - but we should access via the same pattern)

See `helpers/ontology-helper.ts` for the current working implementation.

Basically this task is to tidy that code up into a frontend side ontology lookup service..

DoxasticFox:


application/frontend/src/components/concept-chooser/concept-chooser.tsx
```diff
@@ -4,31 +4,21 @@ import axios from "axios";
...
-      const newCodes = await doBatchLookup("https://onto.prod.umccr.org/fhir", [
-        c,
-      ]);
-      if (newCodes.length > 0 && newCodes[0].display)
-        setDisplay(newCodes[0].display);
+      const newCode = await doLookup(c);
+      if (newCode !== undefined) {
+        c.display = newCode.display;
+        setDisplay(newCode.display);
+      }
```
DoxasticFox: Which page can I access to test that this logic still works?

application/frontend/src/helpers/ontology-helper.ts
```diff
@@ -1,132 +1,245 @@
...
+      resolved: lookupResult.resolved.concat(resolved),
+      unresolved: unresolved,
+    };
   }
 
-  return codes;
+  // Cache the lookup results
+  putManyIntoCache(lookupResult.resolved);
+
+  return lookupResult;
```
DoxasticFox, line 252: Lots of changes in this file, but I'm honestly not sure if most of them are worthwhile. New functions are `putManyIntoCache` and `putIntoCache`. The other functions just had their definitions changed without changing their functionality in a (possibly vain) attempt to improve readability.

application/frontend/src/helpers/ontology-helper.ts
```diff
@@ -1,132 +1,245 @@
...
-  ontoUrl: string,
-  codes: CodingType[],
-  forceAllRefresh: boolean = false
-): Promise<CodingType[]> {
-  // console.log(`Doing a code lookup of ${JSON.stringify(codes)}`);
 
-  if (codes.length === 0) return codes;
+// TODO: read this from config file
+// The URL of the ontology server FHIR endpoint
+const ONTO_URL = "https://onto.prod.umccr.org/fhir";
```
DoxasticFox: Should I config-ize this as part of this ticket?

application/frontend/src/components/concept-chooser/concept-chooser.tsx
andrewpatto: In the Bulk box on any release page is a disease chooser that will trigger this

application/frontend/src/helpers/ontology-helper.ts
andrewpatto: Looks ok - the previous one had very little thought go into it - so any attempt at readability improve is worth it

application/frontend/src/helpers/ontology-helper.ts
andrewpatto: Yep lets do that.

The FHIR ontoserver URL *does* exist in the backend config already - so the only plumbing that needs fixing is the movement of that from backend to frontend..

Which we have a technique for..  see the end of buildSafeEnvironment() in backend.. app.ts

andrewpatto: and EnvRelayProvider and index.tsx in the front end..

That will mean the URL can be passed all the way into a React component via hooks .. but then might need to be passed into ontology helper functions because none of them are React components
(or maybe they should be?)

williamputraintan (approved):

mmalenic (approved):

application/frontend/src/helpers/ontology-helper.ts
DoxasticFox: Thanks

DoxasticFox:
@andrewpatto Could you please take another look?

application/backend/src/app.ts
```diff
@@ -240,6 +240,10 @@ export class App {
     );
     addAttribute("data-build-version", result.build_version || "unknown");
     addAttribute("data-deployed-environment", this.settings.environment);
+    addAttribute(
+      "data-fhir-url",
```
andrewpatto: Sorry - can I be annoying pedantic. There's a possibility that we'll also end up with FHIR servers that are used for retrieving patient data (separate to FHIR terminology - not that you were to know that!). So can we rename throughout something to specifialise this particular type of FHIR url? 
`terminology-fhir-url`

(more specifically it could be `ontoserver-fhir-url` - but given the terminology spec is not ontoserver specific - lets go for the above)

andrewpatto: As in here would be `data-terminlogy-fhir-url`.. but to make that change throughout (again sorry!)

andrewpatto:
Looks good - just as a last check - can you do a search throughout your branch for "onto" and make sure there isn't any "non-config" use of the URL.. I might have a few scattered in places you haven't edited..

application/backend/src/app.ts
DoxasticFox: No need to apologise; I appreciate the pedantry :smile:

DoxasticFox:
@andrewpatto PTAL :pray:

andrewpatto:
excellent thanks - you've cleaned up all the technical debt I managed to produce in (checks notes) two weeks of programming unsupervised..

merge away..

andrewpatto (approved):

PR #78, opened 2022-10-28
Feature/38 redcap dac import

### LINKED ISSUE #38: Import a cardiac flagship Redcap CSV into Elsa as a new 'release'
@sarahcasauria has said we can get an export out of the AG data access control system in some CSV/TSV/JSON format (TBD) and we should give some UI mechanism by which that can be specified and new releases are detected and able to be created in Elsa.

I am considering this UI flow to be like the following (but happy for other thoughts)

A new tab in the import release panel (alongside REMS/DUO) - something like "Australian Genomics Export".

Then on that tab would be a "drag TSV/CSV file here" (any sort of file chooser - but there's heaps of nice react drag-drop/select ones available).

The TSV/CSV would be analysed in react (the file will be small) - and a picker modal dialog displayed (see what happens with the REMS one). The user would then pick which of the ones in the TSV they want to 'create' (we can think about batch in the future - for the moment one by one is ok I think - but feel free to do batch if easy)

The one chosen would then initiate a "Create release" POST(?) API call telling it to fill in a new Release database entry.

The main work here I guess will be determining what is available in the Redcap export and how best it maps across. It may require actual changes to the Release schema. And then working on the specific JSON API schemas needed to support the 'create release' api call.

I wouldn't worry overly about permission structures at the moment - we still need to sort out a role/admin (who can do this operation?) - so we can add those boundaries later. For the moment just go with "everyone logged in" can do this..

andrewpatto:
Finished enough for demonstration purposes
Still need BA work to work out with AG to work out details of the display/mapping


andrewpatto:
This is one that could definitely be improved - but not until we surface more requirements - so submitting this PR in a state that works - but will need a follow up in the future

application/backend/src/api/routes/dac-routes.ts
```diff
@@ -27,15 +33,40 @@
...
+  fastify.post<{
+    Body: AustraliaGenomicsDacRedcap[];
+    Reply: AustraliaGenomicsDacRedcap[];
+  }>("/api/dac/redcap/possible", {}, async function (request, reply) {
+    const { authenticatedUser } = authenticatedRouteOnEntryHelper(request);
+
+    const n = await redcapAgService.detectNewReleases(request.body);
+
+    reply.send(n);
+  });
```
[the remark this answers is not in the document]
andrewpatto: All our routes have a default rate limiter that the codeql never picks up on

andrewpatto:
Resolves #38

application/common/elsa-types/csv-australian-genomics.ts
```diff
@@ -0,0 +1,236 @@
...
+  if (ag.daf_type_research___case_ctrl === "1") return "CC";
+  if (ag.daf_type_research___disease === "1") return "DS";
+  if (ag.daf_type_research___hmb === "1") return "HMB";
+  if (ag.daf_type_research___poa === "1") return "POA";
+  if (ag.daf_type_research___method_dev === "1")
+    throw new Error("No Method Dev");
+  if (ag.daf_type_research___other === "1") throw new Error("No Other yet");
+  if (ag.daf_type_research___popn_stud === "1")
+    throw new Error("No Pop stud yet");
+}
```
DoxasticFox, line 77: Doesn't the type checker say that the return type is missing ` | undefined`?

application/backend/src/business/services/australian-genomics/redcap-import-application-service.ts
```diff
@@ -0,0 +1,198 @@
...
+    const currentReleaseIdentifiers = new Set<string>(
+      await e
+        .select(e.release.Release, (r) => ({
+          applicationDacIdentifier: true,
+          filter: e.op(r.applicationDacIdentifier.system, "=", AG_REDCAP_URL),
+        }))
+        .applicationDacIdentifier.value.run(this.edgeDbClient)
+    );
+
+    const results: any[] = [];
```
DoxasticFox:
```suggestion
    const results: AustraliaGenomicsDacRedcap[] = [];
```

application/backend/src/api/routes/dac-routes.ts
```diff
@@ -27,15 +33,40 @@ export const dacRoutes = async (
 
   fastify.post<{
     Params: { nid: string };
-    Reply: any;
+    Reply: string;
```
DoxasticFox, line 36: :ok_hand:

application/frontend/src/pages/dac-import/australian-genomics-dac-redcap/australian-genomics-dac-redcap-upload-div.tsx
```diff
@@ -0,0 +1,195 @@
...
+  default: {
+    borderColor: GREY,
+  } as CSSProperties,
+  remove: {
+    height: 23,
+    position: "absolute",
+    right: 6,
+    top: 6,
+    width: 23,
+  } as CSSProperties,
```
DoxasticFox, line 81: I see `as` used a lot. The `as` keyword effectively disables the type checker where it's used. A less-unsafe alternative is [type guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards). Or in this case, this stuff could've obviously gone into a CSS file.

DoxasticFox (approved):
No idea what's going on in this PR so I just played "human linter" for a bit then approved.

williamputraintan (approved):

williamputraintan:
LGTM

application/frontend/src/pages/dac-import/australian-genomics-dac-redcap/australian-genomics-dac-redcap-upload-div.tsx
andrewpatto: Yep - it was the demo example from the 'upload' React component and I didn't make any attempt to clean up (or convert to tailwind for instance). The 'upload' component is actually a bit yucky so thinking of pivoting away from it entirely - will put in a ticket for looking at it again later

application/common/elsa-types/csv-australian-genomics.ts
andrewpatto: Yes its undefined in the case where no DUO codes are mentioned.. I'm aborting here because I don't know what behaviour we actually want for these specific ones as the Elsa application model has no support for these yet (its more a come back and fix these kind of error)
There will be lots of future DUO tickets that we need to progress (as in we don't even have the DUO matching algorithm started yet)

PR #80, opened 2022-10-28
Add a aws signing key setting and use it for presinging S3 urls if there

### LINKED ISSUE #73: Create a config/mechanism to use a static AWS user for S3 signing
Generating S3 urls from any 'deployed' AWS infrastructure will give a maximum signing period bounded by the assumed role of the lambda/fargate etc. Even in a case where we maxxed out those role periods - it is not possible for it to be more than 24 hrs (?) - anyhow we wouldn't want to do extend them anyhow.

Existing pre-signed URLs use a static AWS user that can sign things up to 7 days. We need to make mechanism by which those credentials can be made available to Elsa and then used *just* for S3 signing.

andrewpatto:
Waiting a few days to see if the keys are indeed able to last 6 days.. but this should be enough.

### DIFF
--- a/application/backend/jobs/select-job.ts
+++ b/application/backend/jobs/select-job.ts
@@ -25,6 +25,8 @@ const fakeSettings: ElsaSettings = {
   sessionSalt: "",
   sessionSecret: "",
   remsBotUser: "",
+  awsSigningAccessKeyId: "",
+  awsSigningSecretAccessKey: "",
   oidcClientId: "",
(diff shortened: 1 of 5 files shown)


andrewpatto:
for #73

williamputraintan (approved):
LGTM!

andrewpatto:
Confirmed that the pre-signed urls still work after a weekend..

PR #82, opened 2022-10-31
Make a github actions for building a github released Docker build

### LINKED ISSUE #52: Make a github actions for building a github released Docker build
On each 'release' in github - it would be good to have a docker image of Elsa Data built with a corresponding tag. So if we decree a release to be 1.0.1 - we want an elsa-data:1.0.1 built and pushed.

So we need to decide what docker repo this goes into - would prefer something Github related in the first instance (even if we also end up pushing it to some quay.io or something)

The docker image itself I'm not super concerned with (i mean it should build correctly and hopefully be able to be run) - but wouldn't worry too much about all the env variable setup etc we'll need to hook to postgres. This ticket in particular is about the build process that is in github actions.

DoxasticFox:
Resolves https://github.com/umccr/elsa-data/issues/52.

### DIFF
--- a/.github/workflows/docker-image-build.yml
+++ b/.github/workflows/docker-image-build.yml
@@ -1,0 +1,42 @@
+name: Docker Image Build
+
+on:
+  push:
+    branches:
+      - "dev"
+    tags:
+      - '*'
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+      -
+        name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      -
+        name: Login to GHCR
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: andrewpatto
+          password: ${{ secrets.ELSA_DATA_PUBLISH_DOCKER_TOKEN }}
+      -
+        name: Build and push
+        uses: docker/build-push-action@v3
+        with:
+          context: application
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: |
+            ghcr.io/umccr/elsa-data:${{ github.ref_name }}
+            ghcr.io/umccr/elsa-data:${{ github.sha }}
+          build-args: |
+            ELSA_DATA_VERSION=${{ github.ref_name }}
+            GITHUB_SHA=${{ github.sha }}
--- a/README.md
+++ b/README.md
@@ -17,11 +17,11 @@ Then
 
 ```shell
 while IFS= read -r -d '' file; do
-    dirname=$(dirname "$file")
-    (
-      cd "$dirname"
-      npm install
-    ) || exit 1
+  dirname=$(dirname "$file")
+  (
+    cd "$dirname"
+    npm install
+  ) || exit 1
 done < <(find . -name package.json -type f ! -path "*/node_modules/*" -print0)
 ```
 
@@ -34,7 +34,7 @@ npm run build:dev
 3. In folder `application/backend`
 
 ```shell
-edgedb project init
+edgedb project init --non-interactive
 npm run edgetypes
 ```
 
--- a/application/Dockerfile
+++ b/application/Dockerfile
@@ -80,9 +80,6 @@ ENV ELSA_DATA_VERSION $ELSA_DATA_VERSION
 ARG GITHUB_SHA
 ENV GITHUB_SHA $GITHUB_SHA
 
-RUN echo "export ELSA_DATA_VERSION=$ELSA_DATA_VERSION" >> ~/.profile
-RUN echo "export GITHUB_SHA=$GITHUB_SHA" >> ~/.profile
-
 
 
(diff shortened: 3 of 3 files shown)


.github/workflows/docker-image-build.yml
```diff
@@ -0,0 +1,42 @@
...
+          username: andrewpatto
+          password: ${{ secrets.ELSA_DATA_PUBLISH_DOCKER_TOKEN }}
+      -
+        name: Build and push
+        uses: docker/build-push-action@v3
+        with:
+          context: application
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: |
+            ghcr.io/umccr/elsa-data:${{ github.ref_name }}
```
DoxasticFox, line 38: `github.ref_name` is either a branch name (e.g. `dev`) or tag name.

README.md
```diff
@@ -17,11 +17,11 @@ Then
...
-    dirname=$(dirname "$file")
-    (
-      cd "$dirname"
-      npm install
-    ) || exit 1
+  dirname=$(dirname "$file")
+  (
+    cd "$dirname"
+    npm install
+  ) || exit 1
```
DoxasticFox, line 24: Not necessary for this PR but I realised the formatting wasn't right while I was copying/pasting into the Dockerfile.

application/Dockerfile
```diff
@@ -39,30 +39,49 @@ FROM amazonlinux:2022
...
-# the common packages need to be already in place in the folder struct *before* npm install
-COPY common/ ./common/
+RUN dnf install -y nodejs npm findutils shadow util-linux
 
-# install packages for backend and frontend
-COPY backend/package* ./backend/
-RUN cd backend && npm clean-install
-COPY frontend/package* ./frontend/
-RUN cd frontend && npm clean-install
+RUN useradd app-user
```
DoxasticFox, line 47: `edgedb` (specifically `postgres`, which `edgedb` relies on) doesn't like running under the `root` account, which is the default in a Docker container.

application/Dockerfile
```diff
@@ -39,30 +39,49 @@ FROM amazonlinux:2022
...
+
+RUN chown -R app-user /app
+
+SHELL ["/usr/bin/su", "app-user", "-c"]
+
+RUN : \
```
andrewpatto, line 53: Doesn't this mean the running Elsa will always have an Edgedb running as well? This might be the one bit that does need a separate stage-builder (I definitely agree that we will have to install Edgedb and do the migration in order to get the edge types generated - but we then don't need it still in the final image?)

application/Dockerfile
```diff
@@ -39,30 +39,49 @@ FROM amazonlinux:2022
...
 
-# copy the entire backend structure (we can probably pare this back at some point)
-COPY backend/ ./backend/
+RUN : \
+  && cd frontend \
+  && npm run build:dev
 
+RUN : \
+  && cd backend \
+  && export EDGEDB_INSTALL_IN_DOCKER=allow \
```
DoxasticFox, line 73: `EDGEDB_INSTALL_IN_DOCKER` isn't documented; I had to read the [source for edgedb-cli](https://github.com/edgedb/edgedb-cli/blob/e40400c8062a6e8e9ab1580f91fcde0e10f1c385/src/portable/platform.rs#L69-L94) to discover it exists. So I'm not sure how robust this is. There's an official edgedb docker image, but this works for now.

application/Dockerfile
```diff
@@ -39,30 +39,49 @@ FROM amazonlinux:2022
...
+  && npm run edgetypes
 
+ARG ELSA_DATA_VERSION
+ENV ELSA_DATA_VERSION $ELSA_DATA_VERSION
 
+ARG GITHUB_SHA
+ENV GITHUB_SHA $GITHUB_SHA
 
+RUN echo "export ELSA_DATA_VERSION=$ELSA_DATA_VERSION" >> ~/.profile
+RUN echo "export GITHUB_SHA=$GITHUB_SHA" >> ~/.profile
```
andrewpatto: don't ENV variables defined in the Docker already just appear in the environment of the running docker container - without needing them added to .profile?

DoxasticFox: I think they're present during the build, but not in the image. Let me double-check.

application/Dockerfile
DoxasticFox: So should I make this into a multi-stage build with an extra stage which copies (?most of) `/app/frontend` and `/app/backend`?

application/Dockerfile
DoxasticFox: Nope. You were right. Thanks!

application/Dockerfile
andrewpatto: Yeah apologies for the state of the docker build - it was multi-stage with esbuild - and then esbuild was causing various problems so I (short term) changed to a big COPY in one stage and just running `ts-node`.

I think ideally I would like the resulting image to be 'just' the bits that are needed - and happy if that means the Dockerfile has some multi stages and/or very specific COPY statements.

Happy for you to make a judgement call on how much of that you want to do now and how much can be deferred (for instance we need to think about how people would do config folders in custom docker images etc - so all of that might need a bit more planning on the exact folder structure)

andrewpatto (approved):
Happy for you to close this off whenever you feel it is done and we can get some real world usage of it before revisiting

mmalenic (commented):
Looks good!

mmalenic (approved):

PR #84, opened 2022-11-01
Improve release audit logs

### LINKED ISSUE #54: Render release audit logs with a bit more oomph..
The audit logs for a release currently render as

Screen Shot 2022-09-15 at 2 30 12 pm

If we look at audit.edsl - we see that there are some other audit fields recorded but not rendered - things like actionCategory and outcome.

Also the details haven't been dealt with at all.

Would be good to have a stab at improving the 'audit row' display on the frontend (possibly also needs fleshing out the APIs/backend to make sure all the needed data is passed through).

Happy for anything here - but first off would suggest

---

- [ ] Something that indicates the action and outcome..

Maybe

R -> ✅
W -> ? (something to represent types of failures?)

In a column straight after the time?

---

- [ ] Something that gives access to details if asked for

And then maybe a popup at some point that shows the audit details on hover over.. (see what we do for the display of DUO codes in the cases-box - we have an actual separate API call that returns details and we render them in a popup). That means that all the audit details don't need to be returned in every API call.

Or we could decide that really the details of any audit entry are pretty small so it doesn't hurt to return.

---

- [ ] Some type of sensible filtering

Could suggest at the minimum something that restricts to 'only failures' .. or 'only writes/executes' (?)

mmalenic:
Closes #54.

### Changes
* Add more columns to audit log table.
* Time is shown in a 'from now' format, with a tool tip showing the full date on hover.
* Add routes and queries to fetch data for audit logs.
* Drop down details row for audit log details.
* Implement column sorting functionality.
* Add another page to show a full audit log entry.
* Other small changes refactoring bits of code.

### Implementation
* Table uses [React Table v8](https://tanstack.com/table/v8/docs/guide/introduction).
* Data is fetched from the additional routes - `/api/releases/:releaseId/audit-log/details` and `/api/releases/:releaseId/audit-log/:objectId`.

### Improvements that could be made
* Data fetching is inefficient - data is fetched every time a sorting column is clicked without regard for whether that data has already been fetched in a previous query.
* UI could be a little more polished, with themes across the app determined how components are rendered.
* Animations could be used to improve the flow of the UI, such as a [spinning wheel](https://tailwindcss.com/docs/animation#spin) when data is being fetched.
* ~Expander and view audit log entry buttons could be hidden until hovering over a row, as shown [here](https://tailwindcss.com/docs/hover-focus-and-other-states#differentiating-nested-groups).~ I figured this one out! :smile: 
* It could be clearer what the outcome codes mean, perhaps changing the database schema to an enum would make sense here, or using a tool tip.

### Additional features that could be added
* Search functionality, with proper caching for audit entry events, so that they are not repeatedly fetched.
* Refresh button to refresh the table.
* A button to download the entire audit log in a specified file format.
* Render the full audit log entry page as a table for improved readability, at least for the fields where the shape is known.


application/backend/src/api/routes/audit-log-routes.ts
```diff
@@ -3,33 +3,120 @@
...
+      const events = await auditLogService.getEntryDetails(
+        edgeDbClient,
+        authenticatedUser,
+        id,
+        start,
+        end
+      );
+
+      sendResult(reply, events);
+    }
```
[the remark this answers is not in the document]
mmalenic: Should be okay: https://github.com/umccr/elsa-data/pull/78#discussion_r1007583679?

mmalenic:
Maybe it would be better to remove this one and generate it on demand: https://github.com/umccr/elsa-data/blob/5586ecdf9ad04088e291f7849da88e49271cc304/application/backend/src/test-data/mock-json.json

application/backend/src/api/routes/audit-log-routes.ts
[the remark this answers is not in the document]
andrewpatto: Yep - the code scanning is broken

andrewpatto:
Screenshot 2022-11-01 at 4 15 38 pm

Visually

(A) I think the "expand all" probably needs to somehow look different to the "expand row" - I couldn't quite work out what it meant at first
(B) I think we'd probably need to JSON indent format these (I know we don't have any examples of real world usage of the details - but JSON all on one line probably is hard to read)
(C) I think we could possibly do something here .. the internal codes are E,C   0, 4 etc.. but those are meaningless to the users.. maybe some iconography - with "Read", "Execute" etc as tooltop.. and for outcomes I would think something involving Colours and "Success" (in green box say).. and "Error" (in red box)..

application/backend/src/api/routes/release-routes.ts
```diff
@@ -105,17 +102,17 @@ export const releaseRoutes = async (fastify: FastifyInstance, opts: any) => {
...
+      //
+      // const releaseId = request.params.rid;
+      //
+      // const specs: string[] = request.body;
+      //
+      // const setResult = await releasesService.setSelected(
+      //   authenticatedUser,
+      //   releaseId,
+      //   specs
+      // );
```
andrewpatto: I don't think this was meant to be commented out?

application/backend/src/api/routes/release-routes.ts
```diff
@@ -125,17 +122,17 @@ export const releaseRoutes = async (fastify: FastifyInstance, opts: any) => {
...
+      //
+      // const releaseId = request.params.rid;
+      //
+      // const specs: string[] = request.body;
+      //
+      // const unsetResult = await releasesService.setUnselected(
+      //   authenticatedUser,
+      //   releaseId,
+      //   specs
+      // );
```
andrewpatto: same

application/backend/src/business/db/audit-log-queries.ts
```diff
@@ -28,29 +28,91 @@ export const countAuditLogEntriesForReleaseQuery = e.params(
...
+      e.uuid(releaseId)
+    ),
+    order_by: [
+      {
+        expression:
+          orderByProperty === "whoId"
+            ? auditEvent.whoId
+            : orderByProperty === "whoDisplayName"
+            ? auditEvent.whoDisplayName
+            : orderByProperty === "actionCategory"
```
andrewpatto, line 94: not needed to be dealt with now - but I wonder big picture about the usefulness of ordering by some of these fields..

filtering by these field - absolutely .. but ordering by CRUD doesn't seem super useful for a user.

I mean as a general mechanism in this query its fine though..

application/backend/src/api/routes/release-routes.ts
mmalenic: It was giving me warnings about unused variables, but I didn't know if this code was important. I could just remove it.

application/backend/src/business/db/audit-log-queries.ts
mmalenic: Yeah, that's totally fair enough. I think a filtering/search function would probably be more useful.

application/frontend/src/pages/releases/detail/logs-box/logs-box.tsx
```diff
@@ -1,148 +1,448 @@
...
+  };
+};
 
-    return data.map((row, rowIndex) => {
-      const when = parseISO(row.when as string);
+export type AuditEntryTableHeaderProps<TData, TValue> = {
+  header: CoreHeader<TData, TValue> & ColumnSizingHeader;
+};
 
-      // TODO: get timezone from somewhere in config
```
andrewpatto, line 58: How are time zones handled now? Are the tooltip displays in local time and if so what local time zone (of the browser?)

mmalenic:
> (A) I think the "expand all" probably needs to somehow look different to the "expand row" - I couldn't quite work out what it meant at first

Maybe a [double bracket](https://react-icons.github.io/react-icons/search?q=chevrondouble), or one with a [line through it](https://react-icons.github.io/react-icons/search?q=chevronbar)? Otherwise it could also just be text saying "expand all".

> (B) I think we'd probably need to JSON indent format these (I know we don't have any examples of real world usage of the details - but JSON all on one line probably is hard to read)

I couldn't find a way to fetch only part of a JSON from EdgeDB, so I fetched a cutoff string version of it, which may not be a valid JSON. However, this could still be formatted in a prettier way.

> (C) I think we could possibly do something here .. the internal codes are E,C 0, 4 etc.. but those are meaningless to the users.. maybe some iconography - with "Read", "Execute" etc as tooltop.. and for outcomes I would think something involving Colours and "Success" (in green box say).. and "Error" (in red box)..

Totally agree. I would actually advocate for changing the database schema for the audit log so that it returns an enum rather than an error code. That way it's clear right from the beginning. Something like `scalar type OutcomeType extending enum<'Success', 'Minor Failure', 'Serious Failure', 'Major Failure'>;` for example.

andrewpatto:
Yeah I don't know what looks best for the expand all - are there other tables out there on the internet that do this well as an example?

We did have that requirement to try to limit how much comes back of the JSON to the UI.. you could fetch the whole JSON from the db - and then have the backend do the indent/format/line restrictions.. ?? 
i.e DB - (JSON) - backend - (string[]) - UI

The error code schemas I was trying to align with some standard audit codes (hence the weird numbers) - but even if we do change the schema in the db to 'Success' etc - we still have to communicate that down via API to the frontend (and we wouldn't want to send the literal string "Success").

So I think however it is represented in the db - the UI is going to have to do some map of value->display text

application/frontend/src/pages/releases/detail/logs-box/logs-box.tsx
mmalenic: I believe [dayjs](https://day.js.org/docs/en/manipulate/local) should correctly format this to the local time (i.e. the time of the browser) in a [localized format](https://day.js.org/docs/en/display/format#localized-formats), however let me do some testing. It's also possible to set a timezone using some sort of config.

andrewpatto:
I like the detailed log entry page - we will have to do some sort of breadcrumb mechanism so the person can navigate back to the release they were on (but put that in a separate ticket as we possibly need a broader think about breadcrumbs)

application/backend/src/business/db/audit-log-queries.ts
andrewpatto: I think we'll need to decide if it is filtering is doable in the UI as is (i.e. part of the table ordering/header) - or whether it fundamentally needs a "Power Audit" page that might have a really sophisticated audit viewing mechanism

application/backend/src/api/routes/release-routes.ts
andrewpatto: Its the route that allows the Cases checkboxes to work (confirmed that clicking the boxes doesn't work anymore in your branch)..

So don't know why its being warned about..

mmalenic:
> Yeah I don't know what looks best for the expand all - are there other tables out there on the internet that do this well as an example?

The AWS CloudWatch log events table has something similar, although it has a bolder symbol, with separators between the header names. Maybe something that just draws more attention to it would be enough?

> i.e DB - (JSON) - backend - (string[]) - UI

That's a good idea.

application/backend/src/api/routes/release-routes.ts
mmalenic: Ahh, that's my bad then. I thought the functions had no side effects.

application/frontend/src/pages/releases/detail/logs-box/audit-entry-page.tsx
```diff
@@ -0,0 +1,74 @@
...
+type AuditEntryBoxProps = {
+  data?: AuditEntryFullType;
+};
+
+/**
+ * Render the audit entry event.
+ */
+export const AuditEntryBox = ({ data }: AuditEntryBoxProps): JSX.Element => {
+  return (
+    <SyntaxHighlighter
```
williamputraintan, line 66: Just wondering for improvement, I suppose this could be another table view instead of raw JSON

mmalenic: Oh yeah, I see what you mean. I guess the problem is that we don't know the shape of the details JSON object in advanced. But it could just be a table with the known rows and the detail row shown as is?

williamputraintan: Yeah, maybe this could be a note for future improvement. Probably render the key value as it is.

application/frontend/src/pages/releases/detail/logs-box/logs-box.tsx
mmalenic: Can confirm, this follows my operating system's time and time zone. When I set a different time zone it correctly shows the time and date on hover.

mmalenic:
@andrewpatto I've made most of the suggested changes and I think the UI flows a bit better now. I wasn't convinced on the look of any other icon for the 'expand all' button, but I did add a tool tip explaining what it does.

williamputraintan (approved):

andrewpatto (approved):

PR #85, opened 2022-11-01
Complete AG importer (Family Model)

### LINKED ISSUE #66: AG dataset index import mechanism

- [x] Handle file deletion

- [x] Complete the data model

	~ To be discussed with AG

williamputraintan:
- DataCase Id will now use  familyId if given in filename or probandId
- Pedigree Relationship is linked by grouping proband Id taken from the studyId

Resolve partial #66

### DIFF
--- a/application/backend/dbschema/migrations/00001.edgeql
+++ b/application/backend/dbschema/migrations/00001.edgeql
@@ -1,4 +1,4 @@
-CREATE MIGRATION m1mp63vssm4p2d3sceplp3tuv6izp52yq7mkklt2xs3ys34fv3prbq
+CREATE MIGRATION m1riy5sxqwnewpzjthoxvcykrjwxm4qu7kuxvdday5m5eupq3zvpia
     ONTO initial
 {
   CREATE MODULE audit IF NOT EXISTS;
@@ -186,7 +186,7 @@ CREATE MIGRATION m1mp63vssm4p2d3sceplp3tuv6izp52yq7mkklt2xs3ys34fv3prbq
       CREATE OPTIONAL PROPERTY reason -> tuple<system: std::str, value: std::str>;
   };
   ALTER TYPE dataset::DatasetCase {
-      CREATE OPTIONAL LINK pedigree := (pedigree::Pedigree);
+      CREATE OPTIONAL LINK pedigree -> pedigree::Pedigree;
   };
   ALTER TYPE job::Job {
       CREATE REQUIRED LINK forRelease -> release::Release {
--- a/application/backend/dbschema/migrations/00001.edgeql
+++ b/application/backend/dbschema/migrations/00001.edgeql
@@ -1,4 +1,4 @@
-CREATE MIGRATION m1a3afwi5prkoaxvtrblf2zkz2e3pwciql5wiph4vu33rzx75uqxna
+CREATE MIGRATION m14c3ucfvfwnfnpalenmtbjf3zr3mioceo6o6nf55c2yfoqkacuaeq
     ONTO initial
 {
   CREATE MODULE audit IF NOT EXISTS;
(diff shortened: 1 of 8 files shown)


application/backend/dbschema/pedigree.esdl
```diff
@@ -20,15 +23,15 @@ module pedigree {
     }
 
     type PedigreeRelationship {
-        required link individual -> dataset::DatasetPatient;
+        required multi link individual -> dataset::DatasetPatient;
```
williamputraintan: This might be just a workaround/temporary change assigning to `multi` relationship. I am unable to assign the value with a select in typescript without assigning `multi` relationship. Tried on using `filter_single` ensure return one value but it doesn't accept it.

Used in [pedigree-query#44](https://github.com/umccr/elsa-data/blob/c54534359fed556797d315c6969394213b7ed0d6/application/backend/src/business/db/pedigree-queries.ts#L44),  [pedigree-query#64](https://github.com/umccr/elsa-data/blob/c54534359fed556797d315c6969394213b7ed0d6/application/backend/src/business/db/pedigree-queries.ts#L64), [pedigree-query#66](https://github.com/umccr/elsa-data/blob/c54534359fed556797d315c6969394213b7ed0d6/application/backend/src/business/db/pedigree-queries.ts#L66) and so on. 

Docs mention about [this](https://www.edgedb.com/docs/clients/js/update) but I suppose the schema that was used is also `multi` relationship.
```
const actors = e.select(e.Person, ...);
e.update(e.Movie, movie => ({
  filter_single: {title: 'The Eternals'},
  set: {
    actors: actors,
  }
}))
```

application/backend/dbschema/migrations/00001.edgeql
```diff
@@ -171,22 +171,28 @@ CREATE MIGRATION m1mp63vssm4p2d3sceplp3tuv6izp52yq7mkklt2xs3ys34fv3prbq
       CREATE LINK case_ := (.<specimens[IS dataset::DatasetPatient].<patients[IS dataset::DatasetCase]);
       CREATE LINK patient := (.<specimens[IS dataset::DatasetPatient]);
   };
-  CREATE SCALAR TYPE pedigree::KinType EXTENDING enum<isRelativeOf, isBiologicalRelativeOf, isBiologicalParentOf, isSpermDonorOf, isBiologicalSiblingOf, isFullSiblingOf, isMultipleBirthSiblingOf, isParentalSiblingOf, isHalfSiblingOf, isMaternalCousinOf, isPaternalCousinOf>;
+  CREATE SCALAR TYPE pedigree::KinType EXTENDING enum<isRelativeOf, isBiologicalRelativeOf, isBiologicalFather, isBiologicalMother, isSpermDonorOf, isBiologicalSiblingOf, isFullSiblingOf, isMultipleBirthSiblingOf, isParentalSiblingOf, isHalfSiblingOf, isMaternalCousinOf, isPaternalCousinOf>;
```
andrewpatto: I was actually matching these up with the official GA4GH pedigree standard (though had put no thought into it other than glancing at the standard)..
If we had the sex of everyone recorded - the type of biological parent would then be inferrable?

andrewpatto:
This is fantastic - I wasn't even thinking about having pedigrees done properly (I thought you were just going to put them all into the same case as a grouper). But this is a really good step forward on how we'd deal with pedigrees..

application/backend/dbschema/pedigree.esdl
williamputraintan: Possible on using `assert_single`. Trying ...

andrewpatto (approved):

application/backend/dbschema/migrations/00001.edgeql
williamputraintan: I changed to maternal and paternal because I saw the notes [here](https://github.com/umccr/elsa-data/blob/c54534359fed556797d315c6969394213b7ed0d6/application/backend/dbschema/pedigree.esdl#L30), which then I assume it exist. And since you note about maternal, I suppose Paternal is also available 😅

application/backend/dbschema/pedigree.esdl
williamputraintan: It works!

application/backend/dbschema/migrations/00001.edgeql
williamputraintan: > If we had the sex of everyone recorded - the type of biological parent would then be inferrable?

Yes

andrewpatto: you should know that often I don't know what I'm talking about when I write comments

PR #87, opened 2022-11-03
Added enough to demo beacon queries

andrewpatto:
The cohort text box is now saved to the backend
The cohort query is used (extremely limited - sex =/!= and demo beacon genomics query)


application/backend/src/api/routes/release-routes.ts
```diff
@@ -172,16 +172,19 @@ export const releaseRoutes = async (fastify: FastifyInstance, opts: any) => {
     }
   );
 
+  // TODO: this probably should be a Graphql mutate endpoint rather than this hack..
```
andrewpatto, line 172: This.. (but can't reorganise all this before the demo)

andrewpatto:
The code running - here we can see that the query is for males - but with a particular variant (via beacon).

Only Charles ends up selected.

Screenshot 2022-11-03 at 5 28 57 pm

application/backend/src/api/routes/dataset-routes.ts
```diff
@@ -35,7 +35,7 @@ export const datasetRoutes = async (fastify: FastifyInstance, opts: any) => {
         offset
       );
 
-      sendPagedResult(reply, datasetsPagedResult, page, "/api/datasets?");
```
andrewpatto: these were typescript errors uncorrect (the last parameter isn't used anyhow)

williamputraintan (approved):

PR #98, opened 2022-11-08
Allows the secrets to have a leading prefix

andrewpatto:
(otherwise they get named in a way that doesn't distinguish them in an account)


williamputraintan (approved):
LGTM!

DoxasticFox (commented):
I just noticed that line 17 in `remote-migrate.sh` has this:

```bash
echo $PW | edgedb --host elsa-edge-db.dev.umccr.org --port 4000 --user elsa_superuser --password-from-stdin --tls-ca-file $rootfile  migration apply --schema-dir=../../../application/backend/dbschema 
```

If `$PW` or `$rootfile` have spaces, the script will break. But it's not really relevant to this PR, and I know it's a temporary script, so it'll probably be deleted before we ever have a chance to bump into that edge case.

PR #99, opened 2022-11-08
Potential users need to be a thing

andrewpatto:
Because we will be processing DAC applications (that mention users by name, email etc) - but *before* the user has ever interacted with Elsa Data - we need a way to park their user details and reassociate the first time they log in.


andrewpatto:
This is currently being done on 'displayName' - which is incorrect - but waiting for a post demo feature where we will add an email property - at which point that would be the preferred matcher

DoxasticFox (approved):

williamputraintan (approved):

mmalenic (approved):

PR #100, opened 2022-11-08
AG Redcap import tidy up

andrewpatto:
The Redcap import now supports potential users + other tidy ups


application/backend/src/business/services/australian-genomics/README.md
```diff
@@ -47,7 +47,7 @@
 | Phenotypic/Clinical data categories requested (choice=Clinical investigations)                                              | daf_clindata_category\_\_\_4      | 1                                | 0                   |
 | Phenotypic/Clinical data categories requested (choice=Variant information)                                                  | daf_clindata_category\_\_\_5      | 1                                | 0                   |
 | Flagship set requested                                                                                                      | daf_data_req                      | 2                                | 1                   |
-| Describe the subset of participants you are requestingNote: Do NOT input any identifying information into this field        | daf_data_req_subset               | Participants with cardiomyopathy |                     |
+| Describe the subset of participants you are requestingNote: Do NOT input any identifying information into this field        | daf_data_req_subset               | Participants with cardiomyopathy |                     |
```
DoxasticFox, line 50: 🤯 How did this get in here and how did you notice it?

application/backend/src/business/services/australian-genomics/redcap-import-application-service.ts
```diff
@@ -100,9 +115,78 @@ export class RedcapImportApplicationService {
...
+          const r: ApplicationUser = {
+            email: (newApplication as any)[
+              "daf_contact_email_site" + siteNumber
+            ],
+            displayName: (newApplication as any)[
+              "daf_contact_site" + siteNumber
+            ],
+            institution: (newApplication as any)[
+              "daf_institution_site" + siteNumber
+            ],
```
DoxasticFox, line 179: Doesn't the type checker complain that possibly `undefined` values are getting assigned to things which should always be `string`s?

Instead of using `as any`, it'd be safer to define a utility function which gets an object's property if it exists or raises an exception otherwise. Then we'll get something like this:

```suggestion
            institution: getOrRaise(
              newApplication,
              "daf_institution_site" + siteNumber),
```

I was half-expecting `lodash` to have a function like that, but I couldn't see one. (The library follows a functional style, so maybe they're ideologically opposed to making functions not be total by raising exceptions.)

DoxasticFox (approved):

application/backend/src/business/services/australian-genomics/README.md
andrewpatto: Webstorm editor understands markdown tables - and can 'reformat' them so everything lines up.. except if there is a special character everything doesn't line up!

application/backend/src/business/services/australian-genomics/redcap-import-application-service.ts
andrewpatto: Good thinking - I think I possibly need to add a general check on the validity of the resulting ApplicationUser (because we also need to protect against CSV columns that are empty strings for instance). So I'll add that in - and that will pick up any 'undefined' as well)

williamputraintan (approved):

PR #101, opened 2022-11-08
Docker deploy tidy ups

andrewpatto:
When the docker image is actually deployed in Fargate - it had some (small) issues around port numbering (only the root user can bind to port 80)  - so I took the opportunity to move the EdgeDb install into a builder


DoxasticFox (approved):
LGTM

PR #91, opened 2022-11-04
Revisit React upload component used in AG Redcap DAC import

DoxasticFox:
Closes #81. Demo:

[Peek 2022-11-04 20-02.webm](https://user-images.githubusercontent.com/11529449/199934592-61da0c75-2dce-440d-a051-d179b7d610b0.webm)

PR #102, opened 2022-11-08
Random tidy ups I noticed while reviewing deployed demo

andrewpatto:
Nothing much significant in functionality - just visual tidy ups etc

Some temporary that can be reverted back after demos (for instance - we probably *do* want a footer - just no idea what is in the footer yet)


mmalenic (approved):

williamputraintan (approved):

application/frontend/package.json
```diff
@@ -53,11 +53,13 @@
     "react-headless-pagination": "^0.2.1",
     "react-hook-form": "7.33.1",
     "react-icons": "^4.6.0",
-    "react-markdown": "^8.0.3",
+    "react-markdown": "8.0.3",
+    "react-papaparse": "^4.1.0",
```
DoxasticFox: What needs `react-papaparse`? I thought I made this dependency not be required in https://github.com/umccr/elsa-data/pull/91... But, admittedly, I never deleted all the `node_modules` directories after removing it, so I never properly checked if it was still required.

DoxasticFox (approved):

application/frontend/package.json
andrewpatto: sorry that might have been a poor conflict/merge on my part!

PR #105, opened 2022-11-09
#76 - Initial attempt for Egress Data Logs / CloudTrail Logs

### LINKED ISSUE #76: Create a mockup/trial model for how CloudTrail accesses can be rendered inside Elsa
The CloudTrail logs are able to be queried looking for a subset of entries related to each 'release'.

Whilst cloudtrail is excellent for actual audit (like a legal investigation) - the data is probably better situated inside Elsa where it is accessible directly from users/data custodians (who in general won't have access to the AWS backend)

We need to decide on what data we want to bring across (all? some? summarised?) into the Elsa Data db and what level of functionality we want to provide to users.

Some considerations
- log size? We don't want 5 Gib of cloud trail data causing other aspects of Elsa Data to be slow/inefficient..
- UI functionality - what is actually going to be useful to spot outlier behaviour?
- can we make a model that works with all clouds - not AWS specific..

This ticket is just about getting some early thinking - and perhaps a UI mockup of what might be possible to stimulate some thinking..

williamputraintan:
Schema:
- Added a `DataAccessAuditEvent` schema that will hold egress bytes, linking to which FILE is being accessed, and the AuditEvent base .
- ReleaseAudit and DataAccessAudit will have 1:multi relationship, the idea is the release audit that grant the access will hold the `DataAccessAudit` information.

Backend:
- Added `/api/releases/:releaseId/audit-log/:objectId/data-access`, this will have direct access to the `DataAccessAuditEvent` schema
- Added `/api/releases/:releaseId/audit-log/data-access`, this will give a summary of data accessed within this release. It will summarise data from `DataAccessAuditEvent`

Frontend:
- Stealing a lot of code from @mmalenic audit-box  
- Added Data Access Summary release table Screen Shot 2022-11-09 at 2 17 39 pm
- Added Data Access log on the "detailed" audit page Screen Shot 2022-11-09 at 2 19 05 pm

TODO:
- As mentioned in #76, do some plumbing on connecting to AWS for cloudTrailLake. It will also need to have a schema to store eventStoreId for CloudTrailLake querying.

Happy for any comment/feedback for this attempt.


application/frontend/src/pages/releases/detail/logs-box/audit-entry-page.tsx
```diff
@@ -49,6 +50,7 @@ export const AuditEntryPage = (): JSX.Element => {
           {query.isSuccess && <AuditEntryBox data={query.data ?? undefined} />}
         </div>
       </BoxNoPad>
+      <DataAccessLogsBox />
```
andrewpatto, line 53: Happy for this to stay here for the moment - but is this the right spot? The audit entry box above is showing a single audit entry.. 
I'd expect a separate page that lets someone drill down into data access logs..

williamputraintan: Yup, noted. I was thinking that since this single releaseAuditEntry is holding multiple dataAuditEntry, might be the case that this single releaseAuditEntry page will show all associated logs.

application/backend/dbschema/migrations/00002.edgeql
```diff
@@ -0,0 +1,20 @@
...
+    ONTO m1nv55thvouf4q54d2x2fw5m24u7n7vo3ps6vcc674wklv7yoqvvsa
+{
+  CREATE TYPE audit::DataAccessAuditEvent EXTENDING audit::AuditEvent {
+      CREATE REQUIRED LINK file -> storage::File {
+          ON TARGET DELETE ALLOW;
+      };
+      CREATE REQUIRED PROPERTY egressBytes -> std::int64;
+  };
+  ALTER TYPE audit::ReleaseAuditEvent {
+      CREATE MULTI LINK dataAccessAuditEvents -> audit::DataAccessAuditEvent {
```
andrewpatto, line 11: Sorry I'm a bit confused.. I think I thought initially the data access events were tied off a release - but I see they are tied off a Release audit event? What's the rationale for that? Or are you tying them to literally the release event that created the S3 links??

application/frontend/src/pages/releases/detail/logs-box/audit-entry-page.tsx
andrewpatto: yes I think I had a misunderstanding there (see my other comment)

andrewpatto: for your data model I think this UI makes sense (but I probably disagree with the data model!)

andrewpatto:
I think your model might struggle with other forms of data sharing - anything that is continuous access for instance (say a htsget endpoint) might just generate auditable 'data' entries over time - but I don't think would be logically tied to a single parent 'audit event'

andrewpatto (approved):

andrewpatto:
I'm super happy to have this merged - it meets the brief of an initial attempt and is something I'd like to demonstrate.. but will probably need some data model rework at some point

andrewpatto:
@williamputraintan - can you get it to insert a realistic 'sample' audit logs attached whenever it makes *any* release (i see you have test data for the test cases etc). But for this demo - I'd like us to be able to hit the 'download S3 urls' link - and have it immediately have fake entries that we can show to talk about..

application/backend/dbschema/migrations/00002.edgeql
williamputraintan: > Or are you tying them to literally the release event that created the S3 links
Yes, I am trying to link each release event to the DataAccess event.

I changed the s3links to use auditId instead. Not sure if this is too much or not, but I am thinking that this way we could keep track more detailed. 

My thoughts are:
- Release had already a link audit schema, I was thinking to make use of it instead of having multiple audit schema from the release itself. 
- We could track CloudTrailLake record from when it was released, so we could prevent having the recording twice as we could keep track of this cloudTrailLake related to when the link is generated. (Thinking it now, perhaps it is also possible to just query CloudTrailLake based on the releaseId and the last time Elsa had query to CloudTrailLake)

Might just be thinking too complicated 😬

williamputraintan:
Sure, happy for a rework on data-model as it might make more sense now.

DoxasticFox (approved):

PR #107, opened 2022-11-10
Add an npm workspace

mmalenic:
This just makes it a bit more ergonomic to run commands and install dependencies.

For example:
* To install dependencies, `npm install` can be run from the root directory.
* Instead of changing directories in two shell windows, `npm start` can be run from the root directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # elsa-data
 
-Genomic data sharing support software ("let the data go")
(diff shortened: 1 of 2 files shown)


andrewpatto (approved):

williamputraintan (approved):

DoxasticFox (approved):
Nice

PR #110, opened 2022-11-11
First go at meta config system

### LINKED ISSUE #108: Add a 'meta' configuration system that allows configuration settings 'providers' to be selected
We have so far been sourcing configurations based on the entrypoint of invoke..

i.e. docker invokes get one set of settings..
'local' invokes get a different set of settings..

however now we are getting to siutations where (for instance) the Docker image might be invoked locally on a Mac, or on AWS, or on Google - and each of these might need different settings sources.

So we need to introduce a meta configuration (as simple as possible) - that allows the 'deployment' to control the source of settings - not the entrypoint..

Thinking of doing this with a combination of environment variables - and some sensible defaults

andrewpatto:


application/backend/src/config/config-schema.ts
```diff
@@ -22,7 +23,7 @@ convict.addFormat({
 // a TLS artifact that we fetch and possibly save to disk for use in TLS connections
 convict.addFormat({
   name: "tls",
-  validate(val: any, schema: convict.SchemaObj) {
+  validate(val: any, _schema: convict.SchemaObj) {
```
DoxasticFox: What does this change do?

mmalenic: I think it just silences an unused variable warning. However, now I'm doubting myself, because I've been wrong in listening to warnings in the past.

DoxasticFox (approved):

Issue #114, opened 2022-11-14
A super intelligent error display component


andrewpatto:
The errors from our Fastify APIs are generated in RFC 7807 format.

https://www.rfc-editor.org/rfc/rfc7807

It would be good to have an ErrorDisplay component that could take one of these data structures and correctly display whatever information it can (with an understanding of the RFCs rules).

HOWEVER - it would be even better to have an ErrorDisplay component that could handle *any* frontend exception/error - including things like Axios errors where no network connection is made (host not found) - which *aren't* in RFC 7807 format because they never reach our server.

Similarly lots of components just throw Errors (aka the csv parser).

The idea here is not to be perfect everywhere - but to support the common errors we see (i.e. axios).. to support our know RFC 7087 format.. and to default down to displaying plain old error messages if the worst case.

This component has to be super safe though - it should not make *any* assumptions. I'm fine if bad data just displays junk - but the code here shouldn't _fail_ if the RFC 7807 format is invalid (say)

andrewpatto:
The first step here might be a bit of BA work to see in the code what type of errors we currently are catching in the UI frontend and how (poorly) we are handling that

Obviously - if a piece of code can specifically handle the error and deal with it - it will do that first. i.e. a network API call might trap 400 errors or something - but the job of this component is to be the fallback display when the frontend just is saying "I got an error - I need to let the user know"

PR #115, opened 2022-11-14
Fix CSV import for AG

### LINKED ISSUE #112: CSV import for AG doesn't handle trailing end-of-line AND doesn't correctly display exception
When editing my sample CSV with `vi` - it added a trailing EOL character (thereby making a blank line at the end of file).

This causes an error in the CSV parser - though I think it probably should be possible to safely ignore.

Also, the CSV parser only displayed that error in the javascript console - it wasn't displayed as part of the 'uploader' UI

mmalenic:
Fixes #112

### Changes
* Fixes bug which doesn't render an exception box when parsing an invalid CSV file.
* Allows parsing CSV files with empty lines and whitespace-only lines.
* Uses the `ErrorDisplay` component when rendering errors in the `SelectDialogBase` component.

Related to #114 as it improves the `ErrorDisplay` box by introducing red text and background.

### DIFF
--- a/application/frontend/src/app.tsx
+++ b/application/frontend/src/app.tsx
@@ -22,7 +22,18 @@ function NoMatch() {
     <ErrorDisplay
       message={
         <div>
-          No React router match for <code>{location.pathname}</code>
+          <p>
+            No React router match for <code>{location.pathname}</code>
+          </p>
+          <p>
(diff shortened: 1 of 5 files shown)


DoxasticFox (approved):
Thanks for fixing this

williamputraintan (approved):
LGTM

PR #116, opened 2022-11-18
Cookie expiration update

### LINKED ISSUE #96: Frontend can seem to run ok even though login credentials are expired
If you leave the frontend logged in overnight - the next day - you will be able to navigate around and press buttons - but effectively you are not able to do anything because every API request will fail (token expiry).

You are still 'logged in' because the cookie/token still exists - its just no longer useful.

We need to get the frontend to detect this and bounce the user to the login page or something? Not sure how best to handle this - open to options?

- every 5 mins some background javascript that checks the validity of creds and 'logs out' if invalid?
- the frontend detects the specific API results that come about from expired credentials a puts up a message?
-

williamputraintan:


application/frontend/src/providers/logged-in-user-provider.tsx
```diff
@@ -20,7 +21,19 @@ export type LoggedInUser = {
...
 export const LoggedInUserProvider: React.FC<Props> = (props: Props) => {
-  const [cookies] = useCookies<any>([]);
+  const [cookies, setCookie, removeCookie] = useCookies<any>();
+
+  axios.interceptors.response.use(
+    (res) => res,
+    (err) => {
+      const errCode = err.response.status;
+      if (errCode === 403) {
+        alert("Session expired! Please re-logged in back to Elsa.");
```
andrewpatto: I'm not sure about whether we need something this "in your face".. I think its fine for the moment (lets get user feedback first) - but possibly a redirect to the login page is enough

application/backend/src/auth/auth-routes.ts
```diff
@@ -69,6 +69,7 @@ const cookieForUI = (
     path: "/",
     secure: true,
     httpOnly: false,
+    maxAge: 24 * 60 * 60, // 1 day (in seconds)
```
andrewpatto, line 72: I think this is correct for now - but possibly this should scale with the age/expiry that comes back in OIDC tokens (which then leaves the control of session expiry in the hands of the IdPs).

Of course, maybe the IdPs have terrible age/expiry settings - in which case possibly we have some config settings in Elsa that set this.

For the moment leave this as is though..

andrewpatto (approved):

PR #117, opened 2022-11-18
Remove npm workspace

mmalenic:
Perhaps not quite as neat as the workspace solution, however this still allows running `npm install` from the root directory. It should install dependencies correctly to the sub-folders, rather than the root directory.

### DIFF
--- a/application/backend/package-lock.json
+++ b/application/backend/package-lock.json
@@ -11,6 +11,7 @@
       "license": "MIT",
       "dependencies": {
         "@aws-sdk/client-cloudformation": "3.128.0",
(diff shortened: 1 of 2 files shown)


application/backend/package-lock.json
```diff
@@ -11,6 +11,7 @@
       "license": "MIT",
```
mmalenic, line 11: Not sure why this file has changed, I think it's just updating the dependencies?

andrewpatto (approved):
Looks good - well at least to give it a try..

There's some changes coming to the "backend" packages due to the new config system - but we can deal with that after that is merged

PR #120, opened 2022-11-18
issue#97 - Record Email field and PotentialUser fix

### LINKED ISSUE #97: Make email a first class compulsory User property
There are various TODO's throughout the code where we say "should save email address". So whilst we currently save the users

subjectId
displayName

we didn't save email because I didn't know if we should make compulsory. But I think from experience now that we should insist on an email address.

Though we should do this in conjunction with CILogon, where different email addresses can assert the same subject.

Anyhow - we need to work out what to do about emails.

williamputraintan:
- Record `Email` for permission::User from CiLogon attributes
- Change query by `email` instead of temporary `displayName`
- Properly link potentialUser from `startNewRelaese` redcap function
- Some test cases update

### DIFF
--- a/application/backend/src/business/db/user-queries.ts
+++ b/application/backend/src/business/db/user-queries.ts
@@ -26,7 +26,6 @@ export const singleUserByEmailQuery = e.params({ email: e.str }, (params) =>
     .select(e.permission.User, (u) => ({
       ...e.permission.User["*"],
       filter: e.op(params.email, "=", u.email),
-      limit: 1,
     }))
     .assert_single()
 );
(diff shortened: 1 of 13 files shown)


application/backend/src/business/db/user-queries.ts
```diff
@@ -19,47 +19,41 @@ export type SingleUserBySubjectIdType = $infer<
...
-        filter: e.op(params.displayName, "=", u.displayName),
-        limit: 1,
-      }))
-      .assert_single()
+export const singleUserByEmailQuery = e.params({ email: e.str }, (params) =>
+  e
+    .select(e.permission.User, (u) => ({
+      ...e.permission.User["*"],
+      filter: e.op(params.email, "=", u.email),
+      limit: 1,
```
andrewpatto: we can take this limit out for email as it will be unique..
(In fact we want it to fail if somehow there are more than one - which will happen at assert_single())

andrewpatto (approved):

PR #121, opened 2022-11-21
Started on docs and images

andrewpatto:


andrewpatto:
Not finished - but starting approach to documentation (to get us all thinking about this approach)

PR #123, opened 2022-11-22
Feature/testing spike

andrewpatto:
Adding a workflow to github actions that runs tests. In this case also adding Playwright as an e2e testing framework


andrewpatto:
So this is in no way complete - but figure we mind as well get this committed and being used - even if there are not many tests yet.

So this enables 4 levels of testing in github
- a Typescript compile
- unit tests (no dependence on *anything*)
- db tests (needs a running edgeql but nothing else)
- e2e (a UI test suite that runs against the full running solution)

andrewpatto:
I had to make some. minor (non testing) code changes in order for the typescript compile to succeed

application/backend/src/business/services/aws-helper.ts
```diff
@@ -56,14 +57,20 @@ export async function readObjectToStringFromS3Key(
...
-    const bodyBuffer = await getObjOutput.Body.toArray();
+    // https://transang.me/modern-fetch-and-how-to-get-buffer-output-from-aws-sdk-v3-getobjectcommand/
+    const getObjOutput = await s3Client.send(
+      new GetObjectCommand({
+        Bucket: bucketName,
+        Key: objectKey,
+      })
+    );
+    if (getObjOutput.Body) {
+      const stream = getObjOutput.Body as Readable;
```
williamputraintan, line 68: Ahh, thanks for fixing this!

williamputraintan (approved):

e2e/tests/example.spec.ts
```diff
@@ -0,0 +1,18 @@
+import { test, expect } from "@playwright/test";
+
+test("homepage has title and links to intro page", async ({ page }) => {
```
mmalenic, line 3: Cool! Interesting how it tests UI components directly. How specific do you think tests should be? Should we just test the main pages and if all the links work, or should we test things like individual buttons in components?

mmalenic, line 3: Cool! Interesting how it tests UI components directly. How specific do you think tests should be? Should we just test the main pages and if all the links work, or should we test things like individual buttons in components?

mmalenic (approved):

e2e/tests/example.spec.ts
andrewpatto: I'm thinking first point of call is very much 'smoke test'.. basic navigation.. perhaps a single run through of an e2e scenario. But not for instance full regression testing of every component.

DoxasticFox (approved):
Beaut!

e2e/tests/example.spec.ts
andrewpatto: But then we do need to decide on our trade offs and where to best put effort.. should we regression test the React UI components - or is it better to put that effort into really good backend/db/service testing. I think I'd prefer prioritising the latter but we should discuss

PR #125, opened 2022-11-23
Added entry point commands to control startup

### LINKED ISSUE #124: Allow Elsa entrypoint to be given commands other than starting a web server
We have various other tasks that are Elsa related - and share source code with Elsa - but do not actuallly involve running a HTTP web server.

For instance, we need to ability to instruct the db to do a migration. We need to ability to load test data into a demo server.

The way I have seen this done in REMS - is for the invoke/entrypoint to accept a command line of 'commands' - one of which (and the default) starts the web server. But the others can also divert to other code.

andrewpatto:
Part of the path to allowing our Docker image to execute arbitrary admin command as well as being a web server


andrewpatto:
This also is not complete - but need to get this in so that we can deploy the new entrypoints (otherwise even the deployed Elsas will always blank out the database).

This is more a scaffold for how we will do various things (like db migrate) - rather than the actual implementation of them

andrewpatto:
So the idea here is that elsa can be invoked with a sequence of commands (or by default it will just start a web server).

So 

```
node ... entrypoint.ts  db-migrate; db-blank; add-test-scenario 1; start-web-server
```

might be how we setup the Docker CMD for a 'demo' server - so that each time the server is restarted it refreshes.

application/backend/src/entrypoint-command-start-web-server.ts
```diff
@@ -0,0 +1,61 @@
...
+  console.log(`Listening on port ${settings.port}`);
+
+  try {
+    await server.listen({ port: settings.port });
+
+    // TODO detect close() event from the server
+    // TODO possibly replace Bree with our own direct Jobs query and handle that here
+    while (true) {
+      await sleep(5000);
+    }
```
DoxasticFox: What does this do?

DoxasticFox (approved):
Might've been good to use a CLI parsing library but this still LGTM

andrewpatto:
> Might've been good to use a CLI parsing library but this still LGTM

Yeah I would have but most are set up for a single "command" per invoke.. and I really want the ability to have multiples specified with semi colons between.

application/backend/src/entrypoint-command-start-web-server.ts
andrewpatto: I want the web server "command" to actually pause whilst ever the server is running - rather than fall through immediately after start up.

So this is busy wait loop .. it really should be doing a check of the "closing" status of the server..

williamputraintan (approved):

PR #126, opened 2022-11-23
Define watch and build scripts in root package.json

mmalenic:
I forgot to include these in #110. This adds the `dev` and `watch` scripts from the new targets in the backend `package.json`.


andrewpatto (approved):

mmalenic:
Is the `package-lock.json` meant to change so much?

andrewpatto:
Yeah I'm still a bit lost as to what the workspaces is doing for npm.. sometimes it seems to want to be its own node_modules.. and sometimes it leaves it to the subfolders

mmalenic:
We could get rid of it if it's inconvenient. For me, the root `node_modules` folder only has the two dev dependencies listed in the root `package.json`. The rest should be going into the sub-folders.

Although I find the `package-lock.json` a bit confusing..

PR #127, opened 2022-11-23
issue#66 - Datasets Config

### LINKED ISSUE #66: AG dataset index import mechanism

williamputraintan:
- Add Configuration file and schema for elsa's dataset


application/backend/src/entrypoint.ts
```diff
@@ -43,10 +44,20 @@ const start = async () => {
   container.register<ElsaSettings>("Settings", {
     useValue: settings,
   });
-
   await blankTestData();
   await insertTestData(settings);
 
+  // DB Setup
```
williamputraintan: Doubt on putting init dataset records here. I thought would be nice to have all datasets (in config) to be recorded or shown in UI despite of having an empty `DatasetCase`.

Possible to put these on `application/backend/src/entrypoint-command-start-web-server.ts` on [PR125](https://github.com/umccr/elsa-data/pull/125/files). Thoughts?

andrewpatto: Yes - this is a good idea - all the datasets in the config appearing in the UI.

Can we have some 'status' via the API dataset JSON that says whether the corresponding database record exists? (i.e. if noone has even attempted to do a 'load' for the dataset - maybe the UI can indicate this with a "warning" triangle icon or something)

williamputraintan: Do you mean the `DatasetCase`? Could do, yes. I think it is also explicit by seeing the _"0 artifacts ..."_ on the right side of the column in the datasets page.


Screen Shot 2022-11-24 at 10 20 39 am

andrewpatto: Not on DatasetCase - on the actual top level Dataset itself?

I'd contend there is a difference between a dataset that has been attempted to be loaded - and just has 0 artifacts currently - versus one that noone has yet attempted a load.  And 'isLoaded' is a boolean flag indicating whether the Dataset db record exists.

Also, we can add into Dataset a 'lastLoaded' /'lastRefreshed' datetime that is set every time your loader script is run?

application/backend/config/datasets.json5
```diff
@@ -0,0 +1,11 @@
+{
```
andrewpatto, line 1: I would think the datasets config that is actually checked into the application backend would just have our test data sets (10f etc).

A dataset config with cardiac will belong in the CDK deploy setup for UMCCR/AG

andrewpatto (approved):

application/backend/config/datasets.json5
andrewpatto: But happy if that is deferred until we move the CDK into its own repo

PR #128, opened 2022-11-24
Some fix ups to get the Docker deployed elsa working with meta config…

### LINKED ISSUE #108: Add a 'meta' configuration system that allows configuration settings 'providers' to be selected

### LINKED ISSUE #124: Allow Elsa entrypoint to be given commands other than starting a web server

andrewpatto:
… and commands

### DIFF
--- a/application/backend/src/auth/auth-routes.ts
+++ b/application/backend/src/auth/auth-routes.ts
@@ -246,9 +246,22 @@ export const authRoutes = async (
   });
 
   if (opts.includeTestUsers) {
-    const subject1 = await userService.getBySubjectId(TEST_SUBJECT_1);
-    const subject2 = await userService.getBySubjectId(TEST_SUBJECT_2);
-    const subject3 = await userService.getBySubjectId(TEST_SUBJECT_3);
+    // TODO unify this with the other test user insertion code
+    const subject1 = await userService.upsertUserForLogin(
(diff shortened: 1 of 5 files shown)


andrewpatto:
On actual deployment with the new entrypoint and config system - I had to make some quick fixes to get everything working

williamputraintan (approved):

DoxasticFox (approved):

mmalenic (approved):

PR #129, opened 2022-11-24
issue#66 - storage::File `isDeleted` property

### LINKED ISSUE #66: AG dataset index import mechanism

williamputraintan:
[UPDATE 2Dec]

- Add `isDeleted` property in storage::File schema
- Deletion file from AG will now mark storage::File `isDeleted` property to `true`
- Refactoring/renaming AGService in their folder
- Datasets page (in [UI](https://user-images.githubusercontent.com/61998484/205179475-1364133f-81ae-416b-b0db-7f1e838fd4ea.png)) will have an option to show available datasets only or all datasets (included deleted files).
- `/api/datasets/` have a query option to include deleted files (`?includeDeletedFile=true`).
- Minor update on the detailed dataset page on [UI](https://user-images.githubusercontent.com/61998484/205189282-3569e4bc-2343-4e42-8a41-b89bbe086156.png). (Need to do more work on the detailed dataset page but might be on a seperate ticket.)

<details>
<summary>ARCHIVED - 24Nov</summary>

- Add `isAvailable` property in storage::File schema
- Deletion file from AG will now mark storage::File `isAvailable` property to `false`
- Refactoring/renaming AGService in their folder
- Datasets page (in UI) will only show available datasets instead of all datasets. 
- API: 
- `datasets/all` - show datasets summary from all file in db
- `datasets/available` - show datasets summary from only available file

</details>


andrewpatto:
I'm not sure I'd introduce a new API endpoint - I think it would be better to just have a field in the Dataset return values..
(kind of like isAvailable). So isLoaded or something.

Reason being - the 'load' on the server for adding in the available cases is almost zero (its just direct config file -> JSON) - so I think there's very little downside in just always having a /datasets endpoint that always includes all the datasets (and then just flip the boolean based on whether there is a corresponding Dataset Db record or not).

So that means the frontend will always just be calling the single API endpoint - and can do what ever filtering/visuals it wants.

andrewpatto:
Should 'description' be deleted from the Dataset EdgeDb schema - as that is better to be sourced direct from the config?
Are there any other fields that we previously stored in the EdgeDb schema for dataset that in fact could be removed and put into the config only?

andrewpatto:
The renaming to S3Index is good - we should document it at some point - as there is no really well established standard for how to 'set up' files for a dataset. So this approach with manifests etc is as good a defacto standard as any. (and isn't restricted to only AG as you've noted)

application/backend/src/business/db/dataset-queries.ts
```diff
@@ -93,6 +93,140 @@ export const datasetAllSummaryQuery = e.params(
     }))
 );
 
+export const datasetAvailableSummaryQuery = e.params(
```
andrewpatto: Sorry I think I've just realised I've misunderstood the endpoints here??

So the /datasets/available is talking about File availability or dataset availability?

williamputraintan: File availability. So it only takes into account only for files that are available in db. (Deleted file doesn't count).

andrewpatto: By takes into account - you mean in terms of the 'stats' (x artifacts in blah bytes?). That would be the only difference then?

andrewpatto: If so, that's fine - but I don't think I would represent those by two different API endpoints (or at least not named as you have). The 'resource' for `/api/datasets` is 'the dataset'. So I interpret ../available or ../all to mean the availability *of the dataset*.

Instead I think what you are proposing is the ability to alter the meaning of the fields of the resource - so I'd suggest maybe adding a ?param to the endpoint. One default behaviour - and then a parameter '?includeAvailableFileStatsOnly' (or something like that)

application/backend/src/business/db/dataset-queries.ts
```diff
@@ -93,6 +93,140 @@ export const datasetAllSummaryQuery = e.params(
...
+  { limit: e.int32, offset: e.int32 },
```
andrewpatto: I think you could unify the 'all' and 'available' query..

So lets imagine we pass in a boolean parameter into the query..

availabilityMatch: e.bool

application/backend/src/business/db/dataset-queries.ts
```diff
@@ -93,6 +93,140 @@ export const datasetAllSummaryQuery = e.params(
...
+  (params) =>
+    e.select(e.dataset.Dataset, (ds) => {
+      const availBclArtifact = e.select(e.lab.ArtifactBcl, (ab) => ({
+        ...ab["*"],
+        filter: e.op(
+          e.op(ab.bclFile.isAvailable, "=", true),
```
andrewpatto: And then just `e.op(ab.bclFile.isAvaiable, "or", params.availabilityMatch)`

would give you the ability to have one query that does both?

application/backend/dbschema/storage.esdl
```diff
@@ -17,5 +17,12 @@ module storage {
         required property size -> int64;
 
         required property checksums -> array<tuple<type: ChecksumType, value: str>>;
+
+        # This would be useful to identify file that has been deleted by the submitter
+        # We won't be deleting the actual record so it doesn't mess up with any of the linking
+        # Instead a flag to identify if data is still available
+        required property isAvailable -> bool {
```
andrewpatto: The other minor thing I worry about here is that it implies an availability status that we in fact can't guarantee.. i.e the file may in fact *not* be available due to AWS S3 deletion (say). So just literally the naming of the field is possibly too strong.

What we really want to do is mark that we have 'softDeleted' the File (or something along those lines?).

Thoughts?

andrewpatto: An isAvailable flag to me implies that we would be regularly getting Elsa to loop through the File records and actually check that the file is still present or something. (which may not be a bad idea but is not what this feature is about)

andrewpatto:
Can you attach some screenshots of the new datasets table UI

andrewpatto (approved):

application/backend/config/datasets.json5
```diff
@@ -22,12 +22,13 @@
...
       storageUriPrefix: "s3://umccr-10f-data-dev",
     },
-    {
-      name: "10C",
-      description: "UMCCR 10C",
-      uri: "urn:fdc:umccr.org:2022:dataset/10c",
-      storageLocation: "aws-s3",
-      storageUriPrefix: "s3://umccr-10c-data-dev",
-    },
+    // Comment the following to show variety of test-data that is not inclduded in config file
```
andrewpatto: Given that 'dataset not in config' might in the future affect things - do we actually want to make up a completely bogus 4th dataset for this example (10M "UMCCR 10M" the "missing" dataset)?

andrewpatto: The 10C dataset will be an important test dataset for cancer cases (even though it isn't particularly set up yet)

williamputraintan: Yes, sorry. Will add a new dataset then.

williamputraintan:
> Can you attach some screenshots of the new datasets table UI


Screen Shot 2022-12-02 at 1 54 27 pm
(Hovering on the icon led to the tooltip)



Also had some SS from the PR description above on the `UI` word (might need to mentioned this above):
- https://user-images.githubusercontent.com/61998484/205179475-1364133f-81ae-416b-b0db-7f1e838fd4ea.png
- https://user-images.githubusercontent.com/61998484/205189282-3569e4bc-2343-4e42-8a41-b89bbe086156.png

PR #131, opened 2022-11-27
Allow setting host in elsa settings

mmalenic:
The local test users do not work well with an ip address of `0.0.0.0`, so this PR allows setting the host in `ElsaSettings` to another value, such as `127.0.0.1`.


andrewpatto:
As I typed in the "0.0.0.0" into the entrypoint -  I was thinking to myself - this would probably be useful as a setting!

andrewpatto (approved):

PR #132, opened 2022-11-28
Refresh typescript/fastify etc

andrewpatto:
Deal with lots of the compile errors that had worked into our tests Random fixes dealing with updates to packages


andrewpatto:
I'm not sure what the best approach to doing these refreshes is - I just manually bump every few months and then fix the consequences of those package bumps.

Here - we had let our test cases get out of sync with the rest of the code (my fault - I had excludes the tests from the `tsc` config)

So this hasn't made the tests actually work - we still need a round of working on that.

But they at least compile.

DoxasticFox (approved):

PR #133, opened 2022-11-28
Don't paginate search results

### LINKED ISSUE #113: Sample search over the cardiac flagship doesn't seem to be able to match anything
Sample search for 10g and 10f seems to work - but there is something about the (much larger) cardiac flagship that prevents it working.

Guess are:
- it is actually doing the search slowly and just the API calls are timing out silently in the background
- the search actually isn't matching due to the identifiers being different

The search function in the first place was pretty slapdash - so this is probably a good change to revisit now we know more about edgedb queries (maybe we add some indexes?)

DoxasticFox:
Potentially closes #113. Although #113, as written, seems to be describing a different issue that I can't reproduce. The issue I bumped into was instead that, if you navigate to a page other than the first then enter a search query, there won't be any search results. This is because the search box accesses the end point `/api/releases/${releaseId}/cases?page=<some-page>&q=<query-string>`, with `<some-page>` set to the page number at the time the query was entered into the text box. The query string should not contain `page=<some-page>`.

### DIFF
--- a/application/frontend/src/pages/releases/detail/cases-box/cases-box.tsx
+++ b/application/frontend/src/pages/releases/detail/cases-box/cases-box.tsx
@@ -54,10 +54,11 @@ export const CasesBox: React.FC<Props> = ({
     ["releases-cases", currentPage, searchText, releaseId],
     async () => {
       const urlParams = new URLSearchParams();
-      urlParams.append("page", currentPage.toString());
       const useableSearchText = makeUseableSearchText(searchText);
       if (useableSearchText) {
         urlParams.append("q", useableSearchText);
+      } else {
+        urlParams.append("page", currentPage.toString());
       }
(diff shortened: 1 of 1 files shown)


andrewpatto (approved):

DoxasticFox:
@andrewpatto Never thought I'd say this on my own PR, but are you sure you should approve this? :joy: I'm not totally sure if I've fixed the issue, given that #113 is so different to the issue I actually fixed.

andrewpatto:
I mean its definitely a bug - so happy for the fix..

If it turns out not to be the actual bug I encountered - then we can fix that later..

andrewpatto:
I think this is the bug I encountered for #113 - I am guessing that whenever I was playing with search on the cardiac flagship - I must have scrolled through to page 2 or 3 or something - and then done my search.

andrewpatto:
I definitely cannot reproduce the old behaviour on the deployed dev elsa instance

PR #134, opened 2022-11-28
Improve error display

### LINKED ISSUE #114: A super intelligent error display component

mmalenic:
Closes #114 

The follow PR implements an `ErrorBoundary` that can catch any error and display it.

### Changes
* Implement an [`ErrorBoundary`](https://reactjs.org/docs/error-boundaries.html) that can display 7807 errors, axios errors, and generic errors. 
* `ErrorBoundary` can display a message/error without an error being thrown, or when an error is thrown.
* The `ErrorBoundary` is placed at multiple spots in the component hierarchy, and when `useQuery` or `useMutation` hooks are used to interact with the backend.
* Some other small changes and code cleanup.

I think I've gotten most of the logical spots where an `ErrorBoundary` should be, but I might have missed some. Either way, there is an `ErrorBoundary` all the way at the root of the component hierarchy, so an error should never escape without displaying something.

### Recent Changes
* `ErrorBoundary` detects common issues like no internet connection, or an authentication issue. 
* Errors can be displayed in the current `ErrorBoundary`, or re-thrown to a higher up `ErrorBoundary`. This is useful for rendering common errors such as authentication errors in a top-level component rather than multiple times in smaller components.
* There an `EagerErrorBoundary` that display an error passed as a prop, and the regular `ErrorBoundary` which catches thrown errors.
* Many of the components in the `ErrorBoundary` have been rewritten as functional components. 
* Bump prettier version.


mmalenic:
Oh no: https://github.com/umccr/elsa-data/actions/runs/3561901085/jobs/5983201477#step:12:17

I was struggling with this before. Could you guys let me know if it compiles/builds locally for you, because it's working for me currently.

andrewpatto:
Can you put some screenshots of what this looks like in various scenarios.. i.e. when it catches an error all the way at the top level.. and one where it catches an error within a known component. Thanks

andrewpatto:
The compile error (I think) I narrowed down to something to do with inline source maps. Its really frustrating one - and the react scripts are completely useless/unconfigurable/behind the times (almost so much that I wonder if we need to pivot to a different react build chain)

andrewpatto:
https://github.com/facebook/create-react-app/issues/9127

application/frontend/src/pages/users-dashboard/others-box/others-box.tsx
```diff
@@ -87,7 +88,7 @@ export const OthersBox: React.FC<Props> = ({ pageSize }) => {
   };
 
   return (
-    <BoxNoPad heading="Users (not you)">
+    <BoxNoPad heading="Users (not you)" errorMessage={"Something went wrong fetching users."}>
```
williamputraintan: I wonder if we should not wrap `BoxNoPad`/`Box` with `ErrorBoundary`? It seems it doesn't catch the API error and redundant with the error component below (`{dataQuery.isError && ...`)?

application/frontend/src/app.tsx
```diff
@@ -48,35 +49,39 @@ export const App: React.FC = () => {
 
   if (loggedInUser) {
     return (
-      <Routes>
-        <Route path={`/`} element={<HomePage />} />
-        <Route path={`/users`} element={<UsersDashboardPage />} />
-        <Route path={`/dac`} element={<DacImportPage />} />
+      <ErrorBoundary>
```
williamputraintan: I wonder if this should suffice to catch all the thrown error? The Error boundary could still render their children component but have some kind of `Modal` / `Alert` component to show the error? I suppose this should prevent from crashing the whole app?

mmalenic: I think that's a really good idea, I didn't think of that. I think it would be good to have a popup window of some sort with the error, that way, all the errors are presented in the same way, and it doesn't crash the rest of the UI.

application/frontend/src/pages/users-dashboard/others-box/others-box.tsx
mmalenic: I guess my reasoning for having both was just in case something else goes wrong and throws an error, but that's probably unlikely with the `dataQuery.isError && ...` code.

williamputraintan: Hmmm, yeah I guess it will trigger it if there is a rendering error happening inside. I wonder if the `<ErrorBoundary/>` on the Route level would suffice?

application/frontend/src/app.tsx
andrewpatto: I think I'd need convincing the modals are the way to show *all* errors.. I think the majority of "expected" errors would be best dealt with with inline text (near the source of the UI component etc that caused the error?)

I thought modal dialogs are out of fashion with modern UIs?

(I think its fine for completely unexpected errors that might get all the way back to some root component)

application/frontend/src/pages/users-dashboard/others-box/others-box.tsx
mmalenic: It could be, but I think the advantage of having an error boundary closer to a component is that it doesn't render the error on the whole UI, and keeps everything higher up intact.

application/frontend/src/app.tsx
mmalenic: Hmm yeah that's fair. Modals are also a bit harder to implement here as we have to manage the state of dismissing them and closing them, etc. Related: https://github.com/umccr/elsa-data/pull/134#discussion_r1035553302

mmalenic:
Here's some screenshots of the `ErrorBoundary`.

Connection error - still render's the navigation bar:
image

Top level error:
image

Error inside a `Box` component, such as the `LogsBox` table:
image

williamputraintan (approved):

andrewpatto (approved):

andrewpatto:
I'm not 100% sure exactly where this landed - but happy to see it merged and we can finesse as we get some experience for how it works in different situations we encounter

PR #137, opened 2022-11-30
Feature/130 cloud formation as job

### LINKED ISSUE #130: Make the CloudFormation install/remove a proper background job
We have a sharing mechanism that involves creating a custom CloudFormation for an S3 endpoint and installing the cloudformation.

The reverse operation is to remove the CloudFormation.

At the moment both are operating as just long running API calls.. they instead need to be done as proper Jobs.

andrewpatto:
WIP
Doing the CloudFormation creation of S3 access points


andrewpatto:
The Further Restrictions box has been implemented
Backend db rows
API
Frontend tied

Screenshot 2022-12-01 at 5 12 38 pm

Also - any functionality that generates 'files' from the release (like Presigned Urls) will now respect this field when chosing which data types to release.

andrewpatto:
Functionality enabled is that this button can be pressed to install a CloudFormation access point in the background

Screenshot 2022-12-02 at 5 09 54 pm

It then changes to


Screenshot 2022-12-02 at 5 10 11 pm

whilst the install is happening

andrewpatto:
So this is vaguely finished - still some polishing (and for instance the "uninstall" isn't implemented). But I have the demo to AG on the 15th and I'd like this included in the demo - so would be grateful if it could be reviewed and we could look to get it merged even if not perfect.

Then I can nibble at fixes for the little issues that will come as we get to use it

andrewpatto:
The two main "algorithms" - creating the "list of allowed files for release".. and "create the access point cloudformation". I have moved these to separate helper functions and added some basic unit tests for them.

application/backend/src/business/services/jobs/jobs-base-service.ts
```diff
@@ -69,6 +80,196 @@ export class JobsService {
...
+  }
+
+  public async getCloudFormationAccessPointStackForRelease(
+    user: AuthenticatedUser,
+    releaseId: string
+  ): Promise<StackResource | null> {
+    const releaseStackName =
+      AwsAccessPointService.getReleaseStackName(releaseId);
+
+    const releaseReleaseStack = await this.cfnClient.send(
```
williamputraintan: Guessing this should be `releaseStack` instead of `releaseReleaseStack`?

application/backend/src/business/services/jobs/jobs-base-service.ts
```diff
@@ -69,6 +80,196 @@ export class JobsService {
...
+    releaseId: string,
+    s3HttpsUrl: string
+  ): Promise<ReleaseDetailType> {
+    const { userRole } = await doRoleInReleaseCheck(
+      this.usersService,
+      user,
+      releaseId
+    );
+
+    if (userRole != "DataOwner")
```
williamputraintan, line 142: Does this mean only `DataOwner` that can configure AWS access point?

williamputraintan (approved):

application/backend/src/business/services/jobs/jobs-base-service.ts
andrewpatto: I haven't really thought about the permission structures for jobs yet - so was just doing some pretty arbitrary defaults. 

But yes - i think probably (given it includes adding AWS cloudformation) this one will be an admin level activity.

application/backend/src/business/services/jobs/jobs-base-service.ts
```diff
@@ -69,6 +80,196 @@ export class JobsService {
...
+    const releaseStackName =
+      AwsAccessPointService.getReleaseStackName(releaseId);
+
+    const releaseReleaseStack = await this.cfnClient.send(
+      new DescribeStackResourcesCommand({
+        StackName: releaseStackName,
+      })
+    );
+
+    if (!releaseReleaseStack.StackResources) return null;
```
mmalenic: Small change: you could merge this with the below condition. E.g. 
```typescript
if (!releaseReleaseStack.StackResources || releaseReleaseStack.StackResources.length < 1) {
  return null;
}

```

application/backend/src/business/services/jobs/jobs-base-service.ts
```diff
@@ -69,6 +80,196 @@ export class JobsService {
...
+
+    if (releaseReleaseStack.Stacks.length > 1) {
+      throw new Error(
+        "Unexpected result of two cloud formation stacks with the same name"
+      );
+    }
+
+    const theStack = releaseReleaseStack.Stacks[0];
+
+    if (theStack.StackStatus === "CREATE_IN_PROGRESS") {
```
mmalenic: Similar to above, could merge conditions:
```typescript
if (theStack.StackStatus === "CREATE_IN_PROGRESS" || theStack.StackStatus === "DELETE_IN_PROGRESS") {
  return 1;
}
```

mmalenic (approved):

application/backend/src/business/services/jobs/jobs-base-service.ts
williamputraintan: Sure - I am just imagining that researcher would able to do this job (as researcher are the one using it so they should have the control of it) and dataOwner just have the role of approving not configuring for the release?

andrewpatto: Yep - agree that it would be good if researcher can do this - but was initially worried about giving the power to install AWS s3 stuff to the researcher. But maybe we can put enough guard rails around it give the researcher the ability to do this

DoxasticFox (approved):

PR #138, opened 2022-12-01
Move hardcoded URLs into config files

### LINKED ISSUE #61: Move hardcoded URLs into config files
[Here](https://github.com/umccr/elsa-data/pull/60/files#r982776712)'s an example.

DoxasticFox:
Closes #61.


application/backend/src/config/config-constants.ts
```diff
@@ -110,6 +110,50 @@ export const configDefinition = {
     format: "*",
     default: "",
   },
+  mondoSystem: {
```
andrewpatto, line 113: We can/should default these to the actual values (as the definition of snomeds URI is set by SNOMED so is unlikely to change). So if these are set as defaults in here - then *most* setups won't need to define them in their configs at all.

andrewpatto: That said, it is good idea that they are config setting so that if someone *does* want to alter it they can

application/backend/config/base.json5
```diff
@@ -1,4 +1,21 @@
 {
   // an ontoserver with MONDO terminology loaded
   ontoFhirUrl: "https://onto.prod.umccr.org/fhir",
+  mondoSystem: {
```
andrewpatto: see below re: making these the config defaults

andrewpatto (approved):

mmalenic (approved):

PR #148, opened 2022-12-08
Replaced a bunch of backend APIs with a single PATCH

### LINKED ISSUE #140: Replace the Release backend API mutators with a single PATCH op that takes patch instructions
The nature of the UI that we want for releases is that individual UI interactions all get immediately persisted to the backend. So that means we don't CRUD the entire object - we just want to mutate small sections of the object. This is because our 'resources' like release are quite large and complex. We don't really want to do a "save all" of the entire object by POSTing the whole thing to the backend.

Currently these individual mutates have been done ad-hoc - but the routes for release are very ugly. Thinking is to replace the mutates with a single PATCH operation that can handle a sequence of ops.

This ticket is to try this out on the release API and set a pattern for other resources.

andrewpatto:
PATCH operations are driven off a single shared Schema
Some small UI fixes along the way


application/backend/src/api/routes/release-routes.ts
```diff
@@ -169,162 +134,168 @@ export const releaseRoutes = async (fastify: FastifyInstance) => {
...
-              authenticatedUser,
-              releaseId,
-              body.query!
-            )
-          );
-          return;
-        default:
-          reply.status(400).send();
-          return;
+      for (const op of request.body) {
```
andrewpatto, line 165: Whilst this is a big switch - it unifies a lot of boilerplate all into one spot and takes advantage of typescript checking - so each of these case strings *have* to match up with the strings in the corresponding Schema definition.

Also, the Schema is enforced at the route level so the operations submitted *are* strongly type checked before they even get to this route (see the tests I've added in 'integration-tests')

application/common/elsa-types/schemas-release-operations.ts
```diff
@@ -0,0 +1,75 @@
+import { Static, Type } from "@sinclair/typebox";
+import { CodingSchema } from "./schemas-coding";
+import { StringUnion } from "./typebox-helpers";
+
+// “add”, “remove”, “replace”, “move”, “copy” and “test”.
+
+export const ReleasePatchOperationSchema = Type.Union([
```
andrewpatto, line 7: This defines all the operations and types that can be submitted in a PATCH - but is useable as a Schema for checking by the 'route' - but also as typescript classes for the frontend.

application/frontend/src/pages/releases/detail/cases-box/patients-flex-row.tsx
```diff
@@ -39,16 +43,32 @@ export const PatientsFlexRow: React.FC<Props> = ({
 }) => {
   const queryClient = useQueryClient();
 
+  // a mutator that can alter any field set up using our REST PATCH mechanism
```
andrewpatto, line 46: Because we are now talking about typescript safe 'operations' - but all submitted from a single PATCH - we can simplify a lot of the front end mutators

williamputraintan (approved):
LGTM

application/backend/src/api/routes/release-routes.ts
```diff
@@ -169,162 +134,168 @@ export const releaseRoutes = async (fastify: FastifyInstance) => {
     }
   );
 
-  fastify.post<{
+  /**
+   * The main route for altering fields in a release. Normally the UI component for the
+   * field is tied to a mutator which makes the corresponding patch operation.
+   */
+  fastify.patch<{
```
mmalenic, line 141: I think this is a good idea, and a neater way to organise things. Do you think some of the other routes (such as audit logs) would be benefit from a similar restructuring? Maybe that would be part of another PR.

mmalenic (approved):

application/backend/src/api/routes/release-routes.ts
andrewpatto: Yes, I imagine we could do similar for other Resources - but happy to leave to another PR

PR #149, opened 2022-12-08
Refine CloudTrailLake for S3PresignUrl

### LINKED ISSUE #76: Create a mockup/trial model for how CloudTrail accesses can be rendered inside Elsa

williamputraintan:
- Complete mechanism to fetch real data from AWS eventDataStore
- Possible putting this on cron job but could trigger this on-demand by the api layer (`/api/releases/:rid/access-log/sync`). A button is put on the UI.
	Screen Shot 2022-12-08 at 4 48 04 pm
- Add `lastDateTimeDataAccessLogQuery` to release property. To keep track last querying date.


application/backend/src/api/routes/audit-log-routes.ts
```diff
@@ -171,4 +175,73 @@ export const auditLogRoutes = async (fastify: FastifyInstance, _opts: any) => {
...
+    },
+    async function (request, reply) {
+      const { authenticatedUser } = authenticatedRouteOnEntryHelper(request);
+
+      const releaseId = request.params.rid;
+      const accessType = request.body.accessType;
+
+      let replyMessage = "";
+
+      switch (accessType) {
```
williamputraintan, line 212: Just realising when making this PR. We might not need to distinguish between presign/access-point/htsget/etc. I suppose might just run all method (presign/access-point/etc) to query from cloudtrail?

I guess could implement this on later PR when introducing access point data tracking.

williamputraintan: But I think still need to differentiate between cloud provider?

application/backend/src/api/routes/audit-log-routes.ts
```diff
@@ -14,13 +14,15 @@ import {
...
+import { AwsCloudTrailLakeService } from "../../business/services/aws-cloudtrail-lake-service";
+import { DatasetService } from "../../business/services/dataset-service";
 
 export const AuditEventForReleaseQuerySchema = Type.Object({
   page: Type.Optional(Type.Number()),
   orderByProperty: Type.Optional(Type.String()),
   orderAscending: Type.Optional(Type.Boolean()),
 });
-export type AuditEventForReleastQueryType = Static<
+export type AuditEventForReleasQueryType = Static<
```
andrewpatto: this is Releas but probably should be Release

application/backend/src/api/routes/audit-log-routes.ts
andrewpatto: Good catch - but yes wouldn't worry about it for now.. can deal with when we get to our first "other" cloud

application/backend/src/api/routes/audit-log-routes.ts
williamputraintan: Yup, fixing it!

application/backend/config/datasets.json5
```diff
@@ -14,6 +14,9 @@
       uri: "urn:fdc:umccr.org:2022:dataset/10g",
       storageLocation: "aws-s3",
       storageUriPrefix: "s3://umccr-10g-data-dev",
+      aws: {
+        eventDataStoreId: "327383f8-3273-3273-3273-327383f8fc43",
```
andrewpatto, line 18: So the definition of the event data store is per dataset (which makes sense) - but I presume for instance in AG we will just have a single event store defined - and every dataset will point at the same one?

andrewpatto (approved):

application/backend/config/datasets.json5
williamputraintan: Yes I think is likely and possible.

williamputraintan: Well I haven't thought if a dataset could be stored in multiple location (hopefully not 😅)?

PR #155, opened 2022-12-09
Data Cascading

### LINKED ISSUE #94: Look into data ownership relationships in Edge Db to better support cascading delete
I think it would be good if we had clear 'ownership' relationships between objects in Edge Db.

For instance - the Consent statements of a patient belong to a Patient - so don't make much sense if they are retained after a patient is deleted. I believe it is possible to express some of this in the edge db schemas.

I think the end goal is that data can be systematically removed from a db - with broad delete statements. So for instance - if there was a 'bad run' noticed (sequencer machine was contaminated say and this was discovered a month later).

We might say that the operations are the delete/alter any Dataset using those Artifacts.
Then delete the Run itself.
With the expectation that all files uploaded by the run are also cascade deleted.

williamputraintan:
- Add owner relationship in schema
- Add relationship diagram - [LINK](https://raw.githubusercontent.com/umccr/elsa-data/issue-94/docs/architecture/relationship.drawio.svg)
- update DataAccess to copy file record to JSON and remove link to `storage::File` record


application/backend/dbschema/audit.esdl
```diff
@@ -65,7 +65,7 @@ module audit {
         link release_ := .<dataAccessAuditLog[is release::Release];
         
         # Intended file access
-        required link file -> storage::File{
+        link file -> storage::File{
```
williamputraintan: Highlighting this.

Ideally File record shouldn't be deleted at the first place. But if it is deleted, this audit log becomes useless. It no longer have relation to which this audit record is referring to. Alternatively could store this as fileUrl instead of `storage::File` object.

Making this a required object property can cause error when `storage::File` is removed as `audit` record is independent (and shall never be deleted). `dataset::Dataset` owns all record including the `storage::File` so if the Dataset removed file should also be removed? 

The dataset shouldn't removed in the first place, but if necessary at least a Super Admin level could delete from the edgedb instance itself rather than not supporting any dataset deletion?

andrewpatto: Yeah I don't think the audit event should have a strong link to the Edge File object (whether required or not). Audit logs should really have just a copy of the information that was true at the time of the event.

Would there be any problem just copying the file url here?

williamputraintan: Nope, a bit of change to find out the total file size that the log is trying to access. But could just do another query from the fileUrl, and the implementation makes more sense than linking to the file object.

andrewpatto: Feel free to copy the file size as well.. for instance the 'login' and 'logout' audit events I imagine we will put a copy of the display name, subject id and email all into the event as duplicated information

andrewpatto: You can also put this info (file size, url) into the 'details' JSON of the audit event (rather than calling them out as specific data fields).. so each audit event type can be totally in control of what goes into its 'details' object

andrewpatto:
So I think I see the Dataset hierarchy different to the Lab hierarchy. For me, the lab hierarchy is about "creating" data, and the dataset hierarchy is above "organising" data.. and the "release" hierarchy is for releasing it.

So I think if we want to remove data it needs to be removed by the producer (i.e. through the Lab -> Artifact -> File).

And I think if any of that data is "used" by a dataset - then that locks the artifacts in place until they are removed from the dataset (which might be an admin level activity).

So dataset cases->patient->specimens cascade delete down - but stop before artifacts.

But this is just my random thinking and am keen to hear what others think..

andrewpatto:
Apologies - I should have discussed a bit more before this ticket - I realise I hadn't told you the random thoughts in my head

williamputraintan:
Hmmm, so in that case, there are possibility that artifacts exist with no dataset? Would this make confusing that some artifacts records have and some don't have dataset and ideally the file is generated/belongs to a particular dataset?

Well, I think it also make sense of categorising it to producer/organiser/releaser. I guess my thinking was there any benefit of having artifact existing without dataset in the first place?

> Apologies - I should have discussed a bit more before this ticket - I realise I hadn't told you the random thoughts in my head

All good 😁

andrewpatto:
Yes - so things might be indexed in Elsa with no corresponding Dataset.. and you are right - I don't have a _great_ use case for that yet. Possibly more thinking of how Elsa might be an index for *all* artifacts in a lab - not just the obviously shareable ones.

So I think things "belong" to the thing that put them into Elsa i.e the producer. So in the AG case - the producer is nothing more than a file copy from an external lab - but the producer could be our internal pipelines (for instance).

So maybe all our clinical data at UMCCR is indexed in Elsa as it gets produced and goes through clinical processes. But then if someone wants to make a "Dataset" - they can then select artifacts for inclusion (but maybe not all the BAMs.. or only certain types of cases)

andrewpatto:
For me getting "blessed" into a dataset is a higher level of maturity for an artifact.. so yes we might have lots of files, all indexed in elsa - but some of crap quality or data we wouldn't want to release. And then at some point a subset of all the artifacts get put into Datasets which organise them..  and then the final/highest level of maturity is when that artifacts actually gets released out into the wide world.

williamputraintan:
Yup, sounds good!

andrewpatto (approved):

PR #158, opened 2022-12-12
Role in release

### LINKED ISSUE #156: Return users Role in the release as a field in the API
When returning the list of releases, it would be useful if it was customised to the logged in user to show them what sort of role they have (DataOwner etc).

There was a bug in EdgeDb that meant that the query that would enable that data query was failing - but that was 1.x. Should try again until this works (it should be a relatively straight forward edge query so was surprised it failed)

andrewpatto:
The link property "@role" didn't use to be able to be extracted in our edge queries. It is still a bug with Edge but I've just re-oriented the queries to make it work.

### DIFF
--- a/application/backend/src/business/db/release-queries.ts
+++ b/application/backend/src/business/db/release-queries.ts
@@ -1,46 +1,31 @@
 import e from "../../../dbschema/edgeql-js";
 
 /**
- * An EdgeDb query for all the dataset details for those datasets
- * associated with a given release.
- */
-const allReleaseDatasetsQuery = e.params({ releaseId: e.uuid }, (params) =>
-  e.select(e.dataset.Dataset, (ds) => ({
-    ...e.dataset.Dataset["*"],
-    filter: e.op(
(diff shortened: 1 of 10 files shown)


andrewpatto:
Screenshot 2022-12-12 at 4 22 58 pm

Added "Role (in release)" to the UI

andrewpatto:
Also added a Loader div whilst I was there..

andrewpatto:
btw we have a new prettier extension in "pre-commit" that is normalising tailwind css classes (hence why these have automatically changed in the files I've edited).

What I will do soon is do a PR that is specifically just normalising the entire React (but will wait for all front end PRs to be shut so there is no clashes)

andrewpatto:
We are displaying the raw db "PI", "DataOwner" string in the UI - which is something we can revisit (is easy to map to different language strings)

williamputraintan (approved):
LGTM

andrewpatto:
closes #156

PR #161, opened 2022-12-12
Ran pre-commit with -a

andrewpatto:
I have added a tailwind module for prettier and this runs the new modules over all existing source
(it also picked up a couple of random yml files that pre-commit 'fixed')

### DIFF
--- a/.github/workflows/docker-image-build.yml
+++ b/.github/workflows/docker-image-build.yml
@@ -5,17 +5,15 @@ on:
     branches:
       - "dev"
     tags:
-      - '*'
+      - "*"
 
 jobs:
   docker:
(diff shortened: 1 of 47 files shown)


andrewpatto:
Suggestion merging/rebasing this before doing much flowbite work @williamputraintan

andrewpatto:
I this also means everyone should rerun `pre-commit install` after checking this out (to pick up the new pre commit hooks?)

DoxasticFox (approved):

williamputraintan:
> Suggestion merging/rebasing this before doing much flowbite work @williamputraintan

Yup, noted!

PR #166, opened 2022-12-15
backend: add email service

### LINKED ISSUE #154: Introduce a service/config that allows Elsa to send emails
We will need the ability for Elsa to send notification emails - this ticket is about finding a SMTP client component and having it set up so it can be used (in a wrapped NotificationService?)

https://nodemailer.com/about/

Seems to be a nodejs client appropriate

The config of smtpHost etc will need to be set in the config files

mmalenic:
### Changes
* Adds an mail service that allows Elsa to send emails.
    * A mode can be specified to use SES directly, or configure the SMTP server manually.
    * The manual SMTP server takes arbitrary settings according the nodemailer [options](https://nodemailer.com/smtp/#general-options) property in `createTransport`.


application/backend/src/business/services/mail-service.ts
```diff
@@ -0,0 +1,55 @@
...
+    @inject("Settings") private settings: ElsaSettings
+  ) {}
+
+  /**
+   * Setup the mail service.
+   */
+  public setup() {
+    if (this.settings.mailer?.SES?.options != null) {
+      const sesConfig = this.settings.mailer.SES;
+      const ses = new aws.SES({
```
andrewpatto: So this will detect the default SES in the running account?
(are there any special aws permissions needed to do this discovery?)

mmalenic:
I'm not sure what the best way to update this is: https://github.com/umccr/elsa-data/pull/166/files#diff-a586122b4958cc4abd312c743a44ba8ea99f0e987316f979bf21b0056578d3e6.

There is a few new dependencies that seem to cause the tests to fail.

application/backend/src/entrypoint-command-start-web-server.ts
```diff
@@ -8,7 +8,7 @@ import { ElsaSettings } from "./config/elsa-settings";
 import { sleep } from "edgedb/dist/utils";
 import { getFromEnv } from "./entrypoint-command-helper";
 import { DatasetService } from "./business/services/dataset-service";
-import { JobsService } from "./business/services/jobs/jobs-base-service";
+import {MailService} from "./business/services/mail-service";
```
andrewpatto: have you done a pre-commit install in your repo?
(I think prettier if run would have formatted this import differently)

mmalenic: Ahh true, I think I forgot to run that one. It's weird sometimes it doesn't run for me.

andrewpatto: I mean even though I sometimes run prettier manually from Webstorm - at least with the precommit installed it will run the prettier cleanup at least once before checkin

application/backend/src/business/services/mail-service.ts
andrewpatto: (if there are special permsissions needs - just note them here and I can move them onto the Cloudformation stack (I am in the process of moving the cfn out of this repo))

andrewpatto:
> I'm not sure what the best way to update this is: https://github.com/umccr/elsa-data/pull/166/files#diff-a586122b4958cc4abd312c743a44ba8ea99f0e987316f979bf21b0056578d3e6.
> 
> There is a few new dependencies that seem to cause the tests to fail.

So just a rebuild of the package locks with 'npm install' in frontend and backend?

application/backend/src/business/services/mail-service.ts
mmalenic: Yep, this works for me to just fetch the aws settings from env variables. It requires the ability to use a `SendRawEmail` [role](https://nodemailer.com/transports/ses/#example-3).

I'm not sure if there's other permissions that you mean though?

andrewpatto: Oh - ok I get it now.. so this is an alternate path to going through nodemailer? (I was thinking you were going to discover the SMTP settings from AWS and use those to dynamically configure nodemailer)

application/backend/src/business/services/mail-service.ts
```diff
@@ -0,0 +1,56 @@
...
+        ...sesConfig.options,
+        credentialDefaultProvider: defaultProvider,
```
andrewpatto: ??

andrewpatto: There's a standard injection pattern that we use for clients - and maybe can use here? (which. unifies the configs used for making aws clients - see https://github.com/umccr/elsa-data/blob/dev/application/backend/src/bootstrap-dependency-injection.ts)

application/backend/src/business/services/mail-service.ts
mmalenic: Ahh, I haven't enabled it quite like that. There's two options, which is to directly uses an `aws.SES` object with the nodemailer transporter, e.g. https://nodemailer.com/transports/ses/

Or to use the classic nodemailer transporter which configures things like host, port, etc. https://nodemailer.com/smtp/

Both should be usable with SES.

mmalenic: I guess the `aws.SES` path is sort of the one that is auto discovery of settings.

application/backend/src/business/services/mail-service.ts
mmalenic: Ahh I see, that's a good idea.

DoxasticFox (approved):

application/backend/src/bootstrap-dependency-injection.ts
```diff
@@ -25,7 +26,7 @@ export function bootstrapDependencyInjection() {
   // provided here
   // in all deployed AWS this is true
   // for local dev we should set AWS_REGION explicitly when setting shell credentials (aws-vault etc)
-  const awsClientConfig = {};
+  const awsClientConfig = { credentialDefaultProvider: defaultProvider };
```
andrewpatto: In what way doesn't the current config work? (I.e what does setting the default credential provider enable?)

andrewpatto: I tend to use an `granted` or `aws-vault` in the launching shell - so it basically fills in the environment varables of the shell to have aws credentials and they get picked up without any config. And that then mirrors exactly what happens when the code runs online *in* aws fargate or lambda.

mmalenic: It works without the default credential provider. I added it (somewhat aimlessly) to enable different ways of sourcing credentials listed here: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/modules/_aws_sdk_credential_provider_node.html#aws-credential-provider-for-nodejs

It definitely doesn't need to be in though, and if there's no use for it, it's probably better to remove it.

andrewpatto (approved):

PR #167, opened 2022-12-15
Introduced logging config

### LINKED ISSUE #164: Enable setting up of fastify (and global) logging using our configuration mechanism
Fastify will set up pino logging for us - but it has lots of options. We should make some way that we can pass the useful options from our config files into the setup phase of logging in Fastify

andrewpatto:


andrewpatto:
In localhost dev they are now pretty!

Screenshot 2022-12-15 at 12 55 07 pm

andrewpatto:
But more importantly we can set log levels etc in config/env variables

mmalenic (approved):
I was thinking about when we should be using info, warn, etc. I've only been using `console.log()` so far!

DoxasticFox (approved):

andrewpatto:
> I was thinking about when we should be using info, warn, etc. I've only been using `console.log()` so far!

Yes I have another ticket where I will make the same logger available in services - and then we'll need to pivot all our console logs over to real logs. No rush on it though - can do gradually.. the console logs still make it to cloudwatch.

williamputraintan (approved):

PR #168, opened 2022-12-16
Make the logger instance available via DI 

### LINKED ISSUE #165: Enable use of fastify instance logger in services
We should be able to inject the fastify logger instance (created on fastify startup) into the services that are used - thereby unifying all logging throughout the app.

For testing - we can obviously just inject a simpler logger that prints to stdout

andrewpatto:
Start the process of converting console.logs to loggers


andrewpatto:
Haven't done all the conversions - because individual services need to think about logging levels etc (we probably need an overall project discussion of what to log and when). But the base infrastructure is now in place for us to use logging going forward.

DoxasticFox (approved):

mmalenic (approved):

application/backend/src/config/providers/provider-file.ts
```diff
@@ -82,8 +82,6 @@ export class ProviderFile extends ProviderBase {
           const path = resolve(folder, x.name);
           const content = await json5.parse(await fs.readFile(path, "utf-8"));
 
-          console.log(`Loading configuration from JSON5 file ${path}`);
```
mmalenic, line 85: Why get rid of these instead of converting them to use the logger? I guess they were too noisy or not useful enough?

williamputraintan (approved):

application/backend/src/config/providers/provider-file.ts
mmalenic: Ahh my bad. I can see it's being logged here: https://github.com/umccr/elsa-data/pull/168/files#diff-90e5c8f141df1e2b9960caa1a9c8b1002457d9880ad0b58e914da85f601b886bR70-R73. Never mind my previous comment.

PR #170, opened 2022-12-16
Changed the dockerfile to use edgedb images rather than an installer

### LINKED ISSUE #169: Better install of EdgeDb in our docker build so it is somewhat pinned
The current docker build process fetches the edgedb installer from the internet - but is not pinned - and I worry it can drift over time (I.e if we go back to build an older version - the edgedb will be different from what we got on the original build).

There is a technique for using the edgedb docker images specifically as builders - which will pin the version and possibly is more efficient anyway.

andrewpatto:


andrewpatto:
Closing #169

DoxasticFox (approved):

williamputraintan (approved):

mmalenic (approved):

PR #171, opened 2022-12-16
Add a select all/select none to the Cases or Bulk box

### LINKED ISSUE #141: Add a select all/select none to the Cases or Bulk box
Whist the user may sometimes be doing "Bulk" select, there may be other times when they in fact just want to release all of the datasets.. or "none" of the datasets and then select specimens manually.

So we should give them some buttons that *don't* need to do a background Bulk job - and which either add all samples into the selected set or remove the selected set completely.

I think the backend of this is pretty straightforward queries.

I'm not exactly sure where in the front end it goes.. I mean maybe in the actual cases box? Somewhere just above the search?

But then I worry people may think it represents selecting all on just the page?

DoxasticFox:
Closes #141.

### DIFF
--- a/application/backend/src/business/services/release-base-service.ts
+++ b/application/backend/src/business/services/release-base-service.ts
@@ -177,7 +177,7 @@ export abstract class ReleaseBaseService {
    *
    * @param user the user attempting the changes
    * @param releaseId the release id of the release to alter
-   * @param specimenIds the edgedb ids of specimens from datasets of our release
+   * @param specimenIds the edgedb ids of specimens from datasets of our release, or an empty list if the status should be applied to all specimens in the release
    * @param statusToSet the status to set i.e. selected = true means shared, selected = false means not shared
    *
    * TODO: make this work with any node - not just specimen nodes (i.e. setStatus of patient)
@@ -187,18 +187,15 @@ export abstract class ReleaseBaseService {
    */
   protected async setSelectedStatus(
     user: AuthenticatedUser,
+    statusToSet: boolean,
     releaseId: string,
-    specimenIds: string[],
-    statusToSet: boolean
+    specimenIds: string[] = []
   ): Promise<ReleaseDetailType> {
     const { userRole } = await doRoleInReleaseCheck(
       this.usersService,
       user,
       releaseId
     );
-
-    if (specimenIds.length === 0)
-      throw new Error("You must supply some ids to change the status of");
 
     // note this db set we get is likely to be small (bounded by the number of datasets in a release)
     // so we can get away with some use of a edgedb literal 'X in { "A", "B", "C" }'
@@ -220,7 +217,9 @@ export abstract class ReleaseBaseService {
         filter: e.op(
           e.op(s.dataset.id, "in", releaseAllDatasetIdDbSet),
           "and",
-          e.op(s.id, "in", e.set(...specimenIds.map((a) => e.uuid(a))))
+          specimenIds.length === 0
+            ? e.bool(true)
+            : e.op(s.id, "in", e.set(...specimenIds.map((a) => e.uuid(a))))
         ),
       })
     );
@@ -229,7 +228,7 @@ export abstract class ReleaseBaseService {
       this.edgeDbClient
     );
 
-    if (actualSpecimens.length != specimenIds.length)
+    if (specimenIds.length > 0 && actualSpecimens.length != specimenIds.length)
       throw Error(
         "Mismatch between the specimens that we passed in and those that are allowed specimens in this release"
       );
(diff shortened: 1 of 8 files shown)


application/frontend/src/pages/releases/detail/cases-box/cases-box.tsx
```diff
@@ -73,6 +77,37 @@ export const CasesBox: React.FC<Props> = ({
...
+      });
+    } else {
+      releasePatchMutate.mutate({
+        op: "remove",
+        path: "/specimens",
+        value: [],
+      });
+    }
+  };
+
```
DoxasticFox, line 110: This is copypasta from `application/frontend/src/pages/releases/detail/cases-box/patients-flex-row.tsx` with minor changes

application/frontend/src/pages/releases/detail/cases-box/cases-box.tsx
```diff
@@ -73,6 +77,37 @@ export const CasesBox: React.FC<Props> = ({
...
+    axiosPatchOperationMutationFn(`/api/releases/${releaseId}`),
+    {
+      // we want to trigger the refresh of the entire release page
+      // TODO can we optimise this to just invalidate the cases?
+      onSuccess: async () => await queryClient.invalidateQueries(),
+    }
+  );
+
+  const onSelectAllChange = async (ce: React.ChangeEvent<HTMLInputElement>) => {
+    setIsSelectAllIndeterminate(false);
```
DoxasticFox, line 94: The "Select All" checkbox is _not_ indeterminate only immediately after it's clicked and no other checkbox in the `PatientsFlexRow` is clicked. Once a `PatientsFlexRow` checkbox is clicked, the "Select All" checkbox becomes indeterminate until "Select All" is clicked again. This means it's possible for all the cases to be selected or deselected and for the "Select All" checkbox to still display as indeterminate.

To make the behaviour of the "Select All" checkbox more intelligent, a query can be run after each `PatientsFlexRow` checkbox is clicked to check whether all the cases are selected or not. I decided not to do that because the [function to get all the cases](https://github.com/umccr/elsa-data/blob/3fe6da2ce936e29553d90d75e42a3abeb38b28f2/application/backend/src/business/services/release-service.ts#L157) already seemed pretty complicated, and it also seemed like overkill to run a complicated query when it can be avoided.

Does this seem like a sensible design choice, or should I go ahead and query the DB?

andrewpatto: Hmm.. I think its a good design choice for the moment - and I agree with you that it feels like overkill to query the db for that on every change. So lets go with this and see if any users bump on it.

andrewpatto (approved):

application/backend/src/business/services/release-base-service.ts
```diff
@@ -177,7 +177,7 @@ export abstract class ReleaseBaseService {
    *
    * @param user the user attempting the changes
    * @param releaseId the release id of the release to alter
-   * @param specimenIds the edgedb ids of specimens from datasets of our release
+   * @param specimenIds the edgedb ids of specimens from datasets of our release, or an empty list if the status should be applied to all specimens in the release
```
andrewpatto, line 180: I worry *slightly* about the overload here - only because the overload is a bit reversed logic (i.e. empty means all). But its got good comments so that's ok.
I wonder maybe if we should also get a service test for this as it helps reinforce the definition of the API if there are tests directly on this functionality (though I recognise the service tests are in a bit of a shambles at the moment so maybe can be something to do on a later ticket)

DoxasticFox: >I worry _slightly_ about the overload here

Other alternatives are to give `specimenIds` a type like `string[] | undefined`, or `string[] | "all"`. I opted to make the empty list mean "all" because it made the function [total](https://wiki.haskell.org/Partial_functions), but I hesitated for the reason you mention. It'd be nice if we could have a type like `NonEmptyArray<String> | Everything`. I don't have strong options though, and I'll go with whatever you prefer

I'll start working on adding a test to this PR

andrewpatto: I don't have particularly strong opinions either so leave it as is - and we can redo if our opinions firm up.

I'd like to make that function handle all node types (i.e specimens *and* patients *and* cases) - so when we do that functionality we can think about it some more

andrewpatto (approved):

utils/cloudtrail_s3/package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
williamputraintan: I think this file should be removed.

williamputraintan (approved):

PR #172, opened 2022-12-21
S3 AccessPoint CloudTrail

### LINKED ISSUE #150: Add data accessed via Access Point sharing urls into CloudTrail visualisation
We currently find pre-signed URLs in the cloud trail log and report stats on them.

With the access point sharing - we should see cloud trail entries, tagged by the unique access point URL.

We need to process there just like pre-signed URLs

williamputraintan:
- Add CloudTrail Query for object accessed via AccessPoint

closes #150

### DIFF
--- a/application/backend/src/business/services/aws-cloudtrail-lake-service.ts
+++ b/application/backend/src/business/services/aws-cloudtrail-lake-service.ts
@@ -314,7 +314,7 @@ export class AwsCloudTrailLakeService extends AwsBaseService {
             await this.queryAndRecord({
               sqlQueryStatement: sqlQueryStatement,
               eventDataStoreId: edsi,
-              recordDescription: "Accessed via VPC access point.",
+              recordDescription: "Accessed via S3 access point.",
               releaseId: releaseId,
             });
           }
--- a/application/backend/src/business/services/aws-cloudtrail-lake-service.ts
+++ b/application/backend/src/business/services/aws-cloudtrail-lake-service.ts
@@ -193,11 +193,11 @@ export class AwsCloudTrailLakeService extends AwsBaseService {
 
   /**
    * SQL Queries builder
+   * Ref: https://docs.aws.amazon.com/awscloudtrail/latest/userguide/query-limitations.html
    */
   createSQLQueryByReleaseIdReqParams(
     props: CloudTrailInputQueryType & { releaseId: string }
   ): string {
-    // Ref: https://docs.aws.amazon.com/awscloudtrail/latest/userguide/query-limitations.html
     const requestedField =
       "element_at(requestParameters, 'x-releaseId') as releaseId, " +
       "eventTime, " +
@@ -226,7 +226,6 @@ export class AwsCloudTrailLakeService extends AwsBaseService {
       apAlias: string;
     }
   ): string {
-    // Ref: https://docs.aws.amazon.com/awscloudtrail/latest/userguide/query-limitations.html
     const requestedField =
       "eventTime, " +
       "sourceIPAddress, " +
--- a/application/backend/src/business/services/aws-cloudtrail-lake-service.ts
+++ b/application/backend/src/business/services/aws-cloudtrail-lake-service.ts
@@ -12,6 +12,7 @@ import { AwsBaseService } from "./aws-base-service";
 import { AuditLogService } from "./audit-log-service";
 import { ElsaSettings } from "../../config/elsa-settings";
 import { AwsAccessPointService } from "./aws-access-point-service";
+import { Logger } from "pino";
 
 enum CloudTrailQueryType {
   PresignUrl = "PresignUrl",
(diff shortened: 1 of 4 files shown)


application/backend/src/business/services/aws-cloudtrail-lake-service.ts
```diff
@@ -172,11 +221,44 @@ export class AwsCloudTrailLakeService extends AwsBaseService {
...
+  ): string {
+    // Ref: https://docs.aws.amazon.com/awscloudtrail/latest/userguide/query-limitations.html
+    const requestedField =
+      "eventTime, " +
+      "sourceIPAddress, " +
+      "element_at(requestParameters, 'bucketName') as bucketName, " +
+      "element_at(requestParameters, 'key') as key, " +
+      "element_at(additionalEventData, 'bytesTransferredOut') as bytesTransferredOut";
+
+    const conditionStatement = `element_at(requestParameters, 'Host') LIKE '${props.apAlias}%'`;
```
williamputraintan, line 238: A side note here:

There are 2 ways AWS S3 access point (AP) could be accessed. 

One is via the alias given.
e.g. `aws s3 cp s3://AP-NAME-327383f8fc43-s3alias/test.txt a.txt`

One is via access point arn.
e.g. `aws s3 cp s3://arn:aws:s3:ap-southeast-2:012345:accesspoint/AP-NAME/test.txt a.txt`

This query will only works for query via alias as this is the one we share on the manifest, but if somehow researcher would query using the AP arn, it will not match this condition statement.

andrewpatto: Yep fair enough.. I mean I presume the ARN ones appear in cloudtrail - just that we can't easily pick them up?
(if we were indeed looking for a breach in the raw AWS records)

application/backend/src/business/services/aws-cloudtrail-lake-service.ts
```diff
@@ -191,22 +274,57 @@ export class AwsCloudTrailLakeService extends AwsBaseService {
...
+              eventDataStoreId: edsi,
+              apAlias: a,
+            });
+
+            this.logger.debug("SQL statement: ", sqlQueryStatement);
+
+            await this.queryAndRecord({
+              sqlQueryStatement: sqlQueryStatement,
+              eventDataStoreId: edsi,
+              recordDescription: "Accessed via VPC access point.",
```
andrewpatto: technically is an "S3 access point" (the VPC aspect is something extra)

application/backend/src/business/services/aws-cloudtrail-lake-service.ts
williamputraintan: Yes, it appear just have a different format. 

I just think since we don't actually store the AP name and only store the AP alias, we don't have the proper AP name. Not suggesting parsing because when AP name is too long, alias would cut the AP name.

williamputraintan: Adding more notes:

Had tried creating presign based on these 2 value for access point.

1. Creating presign using the AP alias would need to specify s3 endpoint `--endpoint https://s3.ap-southeast-2.amazonaws.com` to make it work.
2. Creating presign based on AP ARN could work directly.

In CloudTrailLake records, both have the same host which is `AP-NAME-012345.s3-accesspoint.ap-southeast-2.amazonaws.com`. This is also have the same value if aws s3 cli is accessing the access point ARN.

Currently we do not save the AP-Name which prevent the CloudTrailLake query to catch this records.

andrewpatto (approved):

application/backend/src/business/services/aws-cloudtrail-lake-service.ts
andrewpatto: I am totally up for us storing as "real" release data (i.e. as first class fields in the release schema) if that helps us track this. So an array of tuples called "specificShares" (?? better name) - and each of those tuples is <str,str>..

<"AP-NAME-012345.s3-accesspoint.ap-southeast-2.amazonaws.com", "Andrew Patterson">

So the name is a record of who created the 'specific share' (if that is useful info?)

We could even roll the master ids associated with pre-signed URLs into the same scheme (i.e. if there is a different sharing id created each time someone clicks "Download Presigned")

Basically - I don't think tracking this is a bad thing - and the tracking info doesn't have to be wedged into the existing audit entry framework - it can be its own record structure that helps you later on create the audit entries

andrewpatto: Or not a tuple - could even be a real EdgeDb object linked via a multi link

williamputraintan: Yep I think also be useful, but would access point be linked to a release instead a person who creates? Say a different person could also use the same access point.

PR #173, opened 2022-12-22
IP Geolocation

### LINKED ISSUE #162: Resolve IP geolocation as part of ETL for cloudtrail logs
It would be useful if the data access audit log also showed the country (and region?) or origin of each IP address.

Rather than defer this to visualisation time - it is probably best to resolve this at ETL phase and just store in the log entry ("AU", "NZ" etc).

williamputraintan:
Added to auto download from maxmind endpoint if `MAXMIND_KEY` is set as environment variable.

Closes #162


application/backend/package.json
```diff
@@ -33,7 +33,9 @@
     "test:unit": "jest --forceExit --detectOpenHandles --runInBand --config jest.unit-tests.config.js",
     "test:integration": "jest --forceExit --detectOpenHandles --runInBand --config jest.integration-tests.config.js",
     "//postinstall": "Give us the ability to locally patch packages after install",
-    "postinstall": "patch-package"
+    "postinstall": "patch-package",
+    "//updateMaxMindDb": "Update MaxMind Database for IP lookup. Set MaxMind key into environment variable 'MAXMIND_KEY'",
+    "updateMaxMindDb": "cd node_modules/geoip-lite && npm run-script updatedb license_key=${MAXMIND_KEY}"
```
williamputraintan: So by default, it has a pre-existing free Db included in the library (which is 3 years old). Eventually, updating this with this command would update the pre-existing file.

Thought on using this pre-existing Db by default, or would rather have it configured with recent data at the time of deploying or none at all.

andrewpatto: I think the issue will be that in general people will use Elsa by sourcing the pre-built Docker image (i.e. docker.io/elsa/2.3.1). And so the question then becomes more what version are we allowed to distribute *inside* that image - my guess is none at all?

andrewpatto (approved):

PR #174, opened 2022-12-23
Update PageSize component

### LINKED ISSUE #153: Add a first cut of a User Profile page reachable from the user dropdown menu
The user dropdown menu that allows users to logout - should add a link here called "Profile" (??) that takes us to a page devoted to settings etc for the logged in user.

The first thing that can move onto this page is the buttons that set the "page size" (see the dashboard page).

For a first cut this will do - more functionality can be added with other tickets.

williamputraintan:
Screen Shot 2022-12-23 at 12 11 51 pm

Closes #153

### DIFF
--- a/application/frontend/src/pages/account-page.tsx
+++ b/application/frontend/src/pages/account-page.tsx
@@ -31,7 +31,7 @@ export default AccountPage;
  * Helper Component
  */
 const PageSizeSetting = () => {
-  const PAGE_SIZE_OPTIONS = [5, 10, 15, 20];
+  const PAGE_SIZE_OPTIONS = [10, 20, 50, 100];
 
   const [cookies, setCookie, removeCookie] = useCookies<any>([
     UI_PAGE_SIZE_COOKIE_NAME,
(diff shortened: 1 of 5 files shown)


application/frontend/src/pages/account-page.tsx
```diff
@@ -0,0 +1,73 @@
...
+  );
+}
+
+export default AccountPage;
+
+/**
+ * Helper Component
+ */
+const PageSizeSetting = () => {
+  const PAGE_SIZE_OPTIONS = [5, 10, 15, 20];
```
andrewpatto: The page size options I put in for the buttons was totally arbitrary - and in part because I wanted to test paging algorithms hence the size 5. Also  I think I just stopped typing at 20..

So possibly we want this array to be more like 10, 20, 50, 100? Thoughts?

andrewpatto (approved):

mmalenic (approved):

application/frontend/src/pages/account-page.tsx
williamputraintan: Yeah agree, doesn't make sense to put 5 as an option. I mean the API itself will return approx the same time anyway, might just display a larger range of data in that case.

PR #180, opened 2023-01-04
Added the Edgedb binary to the docker image

andrewpatto:
Invoke it from db-migrate


andrewpatto:
unfortunately this requires publishing as an image and then redeploy to test - so I'm not actually sure this works..

but anyhow - if someone can pass this through and we'll see..

williamputraintan (approved):

mmalenic (approved):

PR #181, opened 2023-01-04
New config +/- mechanism

### LINKED ISSUE #179: Allow the config files that have lists to support add/subtract
We can currently support any individual config file having the declaration of an array of entries (see here for example our list of datasets). However - the last definition of "datasets" in any config file will win - it will replace all other definitions.

Ideally - config file should be able to accumulate entries across multiple configs (though we should also support "replace" mode).

```
{
  datasets: [
    {
      name: "Cardiac",
      description: "The Australian Genomics Cardiac flagship",
      uri: "urn:fdc:australiangenomics.org.au:2022:dataset/cardiac",
      storageLocation: "aws-s3",
      storageUriPrefix: "s3://agha-gdr-store-2.0/Cardiac",
    },
}
```

So ideally - we should be able to have another config file

```
{
  "+datasets": [
    {
      name: "Diabetes",
      description: "The Australian Genomics Diabetes flagship",
      uri: "urn:fdc:australiangenomics.org.au:2022:dataset/diabetes",
      storageLocation: "aws-s3",
      storageUriPrefix: "s3://agha-gdr-store-2.0/Diabetes",
    },
```

indicating via the "+datasets" that instead here we want to not replace the existing entries - but add to them.

Similarly - we could imagine a syntax like

```
{
  "-datasets": [
      "urn:fdc:australiangenomics.org.au:2022:dataset/diabetes"
 ]
}
```

which would allow a config file to remove previously encountered entries.

andrewpatto:


andrewpatto:
Still need to add a unit test

andrewpatto:
Closes #179

andrewpatto:
This PR enables syntax in config files such as

```
{
  // use the + notation to add into the existing datasets
  "+datasets": [
    {
      name: "datasetadd1",
      description: "A dataset sample added as 1",
      uri: "urn:elsa.net:2022:datasetadd1",
      storageLocation: "aws-s3",
      storageUriPrefix: "example/datasetadd1",
    },
    {
      name: "datasetadd2",
      description: "A dataset sample added as 2",
      uri: "urn:elsa.net:2022:datasetadd2",
      storageLocation: "aws-s3",
      storageUriPrefix: "example/datasetadd2",
    },
  ],
  // and use the - notation to remove an existing dataset (by name but could also be by uri or id)
  "-datasets": ["dataset001"],
}
```

application/backend/src/config/config-schema.ts
```diff
@@ -40,6 +40,89 @@ convict.addFormat({
...
+          arrayProcessed[am].push(...newStrippedConfig[addKey]);
+
+          // strip out the +?? key as we have handled it and we don't want to pass it on to convict (which won't understand it)
+          delete newStrippedConfig[addKey];
+        }
+
+        if (deleteKey in newStrippedConfig) {
+          for (const toDelete of newStrippedConfig[deleteKey]) {
+            arrayProcessed[am] = arrayProcessed[am].filter(
+              (v) => !idPresent(toDelete, v)
```
mmalenic, line 108: Should there be a warning if there is an attempt to delete a non-existent dataset/super admin? This might be a little bit more user friendly if there is a typo. Possibly there could be a test case for this too.

mmalenic (approved):

application/backend/src/config/config-schema.ts
andrewpatto: Yes - was tossing this up - from it being a fatal error, to it being a warning - all the way to it being ignored! I couldn't really decide what behaviour I'd expect as a user.

Certainly typos are a possibility (which would be good to trap) - but then equally likely I think is someone who sets up their config in a sensible way (i.e remove all the test datasets) - and then years later, someone changes the test datasets config. At that point what do I think as a user? Do I want to be alerted or do I not care?

andrewpatto: I'll go with the warning for now - its a sensible middle ground

andrewpatto: actually - there is no logger available during the config setup - so I'm going for a hard error so that it is definitely noticed. We'll see if it trips people up too much

PR #184, opened 2023-01-10
Introduce paging into the "cases" search box results

### LINKED ISSUE #147: Introduce paging into the "cases" search box results
Currently the UI layout (and APIs) for doing searches in the cases box - are predicated on the assumption that the search results will be minimal and hence don't need paging.

We probably should instead plan for a world where multiple pages of search results come back.

DoxasticFox:
Closes #147. Reverts changes in #133.

### DIFF
--- a/application/frontend/src/components/box-paginator.tsx
+++ b/application/frontend/src/components/box-paginator.tsx
@@ -36,7 +36,10 @@ type Props = {
 export const BoxPaginator: React.FC<Props> = (props) => {
   // for our pagination buttons and text we need to do some page calculations
   const maxPage = Math.ceil(props.rowCount / props.rowsPerPage);
-  const from = (props.currentPage - 1) * props.rowsPerPage + 1;
+  const from = Math.min(
+    (props.currentPage - 1) * props.rowsPerPage + 1,
+    props.rowCount
+  );
   const to = Math.min(
     (props.currentPage - 1) * props.rowsPerPage + props.rowsPerPage,
     props.rowCount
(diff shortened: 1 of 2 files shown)


application/frontend/src/components/box-paginator.tsx
```diff
@@ -15,14 +15,15 @@ type Props = {
 
   // if present, enables a replacement of the paging mechanism for when there is search text
   currentSearchText?: string;
-  setSearchText?: (text: string) => void;
-  clearSearchText?: () => void;
```
DoxasticFox, line 19: Removing this line wasn't needed for this PR, but I noticed that `clearSearchText` didn't seem to be used

application/frontend/src/components/box-paginator.tsx
```diff
@@ -69,19 +62,26 @@ export const BoxPaginator: React.FC<Props> = (props) => {
...
+              className={classNames("text-left text-sm text-gray-700", {
+                invisible: props.isLoading === true,
+              })}
+            >
+              Showing <span className="font-medium">{from}</span> to{" "}
+              <span className="font-medium">{to}</span> of{" "}
+              <span className="font-medium">{props.rowCount}</span>{" "}
+              {props.rowWord}
+            </p>
+          </div>
```
DoxasticFox, line 85: tl;dr: Keep the old "Showing _x_ to _y_ of _z_ cases" message as-is. Except, when `props.isLoading === true`, turn it into a [skeleton](https://flowbite.com/docs/components/skeleton/#default-skeleton) by setting its `visibility` to `hidden` and doing some magic (i.e. the classes `animate-pulse rounded-full bg-gray-200`) to its container. I added `min-w-[200px]` to the container so that it doesn't jump around if the length of the "showing ... cases" message changes.

I did the same kind of thing for the pagination buttons too.

andrewpatto: nice!

andrewpatto (approved):

PR #191, opened 2023-01-25
Handle joint VCF

### LINKED ISSUE #187: Convert the 10f and cardiac loaders to support trio VCFs in shared files
We need to support the ability of a VCF file with 3 subjects in it - to be shared as a single file - but pointed to from 3 different specimens.

The first example of this is in the 10f test data set - where we can see for one of the families we need to create "incorrect" made up paths to the trio VCFS in order that the File records are unique. The correct thing would be to have a single File record for the combined VCF - and we need to store somewhere which "sample id" is then used to fetch the data from the VCF.

This ticket is just about fixing the database modelling/loader for trio VCFs - I don't think this will actually affect any other aspects of the system - but we will make some other tickets regarding how to correctly share the trio files.

As part of this - can we also fix the cardiac loader.

williamputraintan:
- Schema for `lab::ArtifactBase` include an array of `sampleIds`:`string[]` that will be used as a placed holder on what the artifact content.
- AG import mechanism to deal with multi-studyId

#187


application/backend/dbschema/lab.esdl
```diff
@@ -1,6 +1,7 @@
 module lab {
 
     abstract type ArtifactBase  {
+        property sampleIds -> array<str>;
```
andrewpatto: What's the rationale for sampleIds to be a property on all artifacts and not just ArtifactVCF?

williamputraintan: Well I just thought that all artifact contain at least a sampleId and was thinking just to be fair with other artifacts that you could see sampleIds from the lab property. (and all of them have the same relationship to the sampleId anyway).

If this doesn't make sense happy to make it available just for ArtifactVCF.

andrewpatto: I mean I like the thinking - but for instance - some files *don't* have sample ids.. (fastqs have sets of sequencer ids?).. not sure BCLs have anything at all. And BAMs have a sample id - but explicitly only one. I think I'd prefer that we make the Artifact* objects correctly model whatever it is that they wrap. So sampleIds array for VCF.. and just plain sampleId for BAM.. and the others nothing at all

andrewpatto: And in the future we can start to extend this out to whatever we want.. so we might decide to parse elements of the VCF header and store them in ArtifactVCF (list of contigs? reference genome?)

And we could do similar exercises with each of the other formats too

williamputraintan: Ahhhh, OK, will move it out from the ArtifactBase.

application/backend/src/test-data/test-data-helpers.ts
```diff
@@ -335,16 +336,27 @@ export async function createArtifacts(
...
-            checksums: vcfIndex.checksums,
-          }),
+          sampleIds: vcfSampleIds,
+          vcfFile: e
+            .insert(e.storage.File, {
+              url: vcf.url,
+              size: vcf.size,
+              checksums: vcf.checksums,
+            })
+            .unlessConflict((file) => ({
```
andrewpatto, line 346: So this becomes a select (of the existing entry) if it conflicts?

application/backend/tests/service-tests/ag-s3-import.test.ts
```diff
@@ -40,7 +51,7 @@ import {
 } from "../../src/business/db/helper";
 import { registerTypes } from "./setup";
 import { DatasetService } from "../../src/business/services/dataset-service";
-
+import { f } from "edgedb/dist/reflection/builders";
```
andrewpatto: ??

andrewpatto (approved):

application/backend/src/test-data/test-data-helpers.ts
williamputraintan: Yes

Ref: https://www.edgedb.com/tutorial/data-mutations/upsert/select-or-insert

PR #192, opened 2023-01-25
A dialog that allows the creation of 'releases' in a completely manual way

### LINKED ISSUE #79: A dialog that allows the creation of 'releases' in a completely manual way
Currently the only way to create releases is via an upstream DAC (REMS, AG Redcap).

It is possible that people may have workflows involving decisions via email (or face to face) etc.. and then a release might need to be made manually.

I imagine this is a dialog that allows
- selection of datasets
- input of basic details like title
- ability to say who the PIs etc are

And then just goes and makes a release corresponding to this

DoxasticFox:
### Demo

[simplescreenrecorder-2023-01-25_22.47.59.webm](https://user-images.githubusercontent.com/11529449/214556122-e9c5cf02-f801-4261-aca1-f0aa0aaf24a0.webm)

#### General Notes and Questions:

* In the demo, you can see that no cases are visible. This is because I give the release-creating user a "Member" role. Should this be something else, like "DataOwner"?
* I changed the appearance of the vertical tabs because I found the previous appearance confusing :sweat_smile:
* I remember @andrewpatto said the dialog should ask for PIs' email addresses, but this discussion says to prefer the term "applicant" or "data recipient" (which is more consistent with the rest of Elsa): https://github.com/umccr/elsa-data/discussions/35
* I might be confounding a _release_ with an _application_, both in the UI and in my variable names, so I'd be glad to hear feedback on that

This PR closes #79.


application/backend/src/business/services/release-service.ts
```diff
@@ -101,6 +105,81 @@ export class ReleaseService extends ReleaseBaseService {
...
+        set: {
+          releaseParticipant: {
+            "+=": e.select(e.release.Release, (r) => ({
+              filter: e.op(e.uuid(releaseRow.id), "=", r.id),
+              "@role": e.str("Member"),
+            })),
+          },
+        },
+      }))
+      .run(this.edgeDbClient);
```
andrewpatto, line 178: We probably need to handle the situation where the user/emails being invited into the release have not yet logged into Elsa (so don't exist yet as a User - but instead exist as a PossibleUser (sorry can't look up the exact name))

andrewpatto (approved):

andrewpatto:
Overall PR looks spot on to me. Only issue outstanding is the one about PossibleUser v real User. 

And there is a bit of work we need to do on user permission modelling (i.e. DataOwner v PI etc) - but I think we need a bigger picture discussion on that with AG - so wouldn't hold up this PR.

application/frontend/src/pages/dac-import/manual/releases-dashboard-manual-entry-dialog.tsx
```diff
@@ -0,0 +1,227 @@
...
+            >
+              Cancel
+            </button>{" "}
+          </>
+        }
+        content={
+          <>
+            <div className="prose mt-2">
+              <p className="text-sm text-gray-500">
+                This form should only be used in the case where an application
```
andrewpatto, line 112: maybe "application for data could not.."

application/frontend/src/pages/dac-import/manual/releases-dashboard-manual-entry-dialog.tsx
```diff
@@ -0,0 +1,227 @@
...
+                could not be imported via an upstream DAC such as REMS or
+                REDCap.
+              </p>
+            </div>
+            <div>
+              <input
+                type="text"
+                className="my-4 block w-full rounded-lg border border-gray-300 text-sm text-gray-900 focus:border-blue-500 focus:ring-blue-500"
+                placeholder="Release Title"
```
andrewpatto, line 121: So I think we should unify this language to what is used upstream in real DACs.. so possibly "Project Title" or something. But don't need to hold this PR about that language.. we can unify later

application/frontend/src/pages/dac-import/manual/releases-dashboard-manual-entry-dialog.tsx
```diff
@@ -0,0 +1,227 @@
...
+                      name="studyType"
+                      label="Clinical Care Use"
+                      value="CC"
+                      onChange={(e) => onChange(e.target.value)}
+                    />
+                  </RhRadios>
+                )}
+              />
+              {errors.studyType && <Required />}
+              <RhChecks label="Assertions" className="my-4">
```
andrewpatto: Whilst the "study type" radio button is useful because it really is mandatory - I'm not sure I'd duplicate these checkboxes here. Eventually there might be 10(?) checkboxes that are editable as part of the Bulk editing in the release - and we probably don't need them duplicated here. The same user who initiates this manual creation can easily fill them in later (I mean its not wrong to put them here, but we can probably skip it and save ourselves some hassle)

application/frontend/src/pages/dac-import/manual/releases-dashboard-manual-entry-dialog.tsx
DoxasticFox: I raised an issue: https://github.com/umccr/elsa-data/issues/193

PR #198, opened 2023-02-07
Google Storage URL presigner

### LINKED ISSUE #194: Preliminary implementation/exploration of create a GS signed link creator (ala S3 signer)
We should make a release mechanism that signs GS object links in a manner similar to the way we generate S3 links.

Can be pretty rough and ready at the moment - but should investigate things like
- google creds needed (if running inside Google they might be inferred from the env just like AWS?)
- signing mechanism
- some test data (ala some gs://test-data/my.bam sitting somewhere)
- a panel on the releases page that can generate a CSV with GS signed objects

Again, all just learning experience level code - happy for this to then turn into other tickets for finishing/productionisation. I don't know much about GCloud which is why we probably need to start with this kind of investigation ticket.

DoxasticFox:
Closes #194. Demo:

[simplescreenrecorder-2023-02-07_15.21.23.webm](https://user-images.githubusercontent.com/11529449/217147815-e0eac0dc-ca8d-46c2-b3e3-b887c3788e4a.webm)


application/backend/src/business/services/_release-file-list-helper.ts
```diff
@@ -54,6 +58,92 @@ export async function createReleaseFileList(
     return "";
   };
 
+  // note that we are processing here every artifact that is accessible from
+  // the chosen specimens
+  // below we apply some simple logic that will rule out classes of artifacts
+  // based on the isAllowedReadData, isAllowedVariantData etc
+  const unpackFileArtifact = (
```
DoxasticFox, line 65: This function is just code copied/pasted out of the loop below with some slight changes. I did this to help the type system infer the types.

application/backend/src/business/services/_release-file-list-helper.ts
```diff
@@ -62,92 +152,72 @@ export async function createReleaseFileList(
...
-      } else {
+      const unpacked = unpackFileArtifact(fa);
+      if (unpacked === null) {
         continue;
       }
 
-      // decompose the S3 url into bucket and key
-      const _match = entry.s3Url.match(/^s3?:\/\/([^\/]+)\/?(.*?)$/);
+      // decompose the object URL into protocol, bucket and key
+      const match = unpacked.objectStoreUrl.match(
```
DoxasticFox, line 161: Should I have named this `_match`?

application/backend/src/business/services/_release-file-list-helper.ts
```diff
@@ -62,92 +152,72 @@ export async function createReleaseFileList(
...
+        /^([^:]+):\/\/([^\/]+)\/(.+)$/
```
DoxasticFox, line 162: The previous regex implied that the scheme/protocol could be `s3://` or `s://`. Was that intentional? I've only seen `s3` used. If `s` is another possibility, I need to make some minor changes to this PR. Specifically, I'd change `IPresignedUrlProvider` so that its `protocol` field is a list of strings. The `protocol` field is how a class which implements `IPresignedUrlProvider` knows that it should handle a particular URI.

application/backend/src/business/services/gcp-presigned-urls-service.ts
```diff
@@ -0,0 +1,52 @@
...
+  async presign(
+    releaseId: string,
+    bucket: string,
+    key: string
+  ): Promise<string> {
+    const options: GetSignedUrlConfig = {
+      version: "v4",
+      action: "read",
+      // 7 days (minus 5 minutes to avoid bumping into Google's time limit)
+      expires: Date.now() + 1000 * 60 * 60 * 24 * 7 - 1000 * 60 * 5,
```
DoxasticFox, line 41: GCP supports expiry times up to 7 days. Unlike in AWS, whose expiry is a relative time, the time you pass to GCP is absolute. If you specify a time more than 7 days in the future, the `presigned` function will fail. Subtracting 5 minutes is a workaround for this. Of course, we assume the clock on Elsa's server is running no more than 5 minutes early.

An alternative is to use Google's v2 API, which supports longer expiry times. Unfortunately, it's a legacy API.

application/backend/src/business/services/presigned-urls-service.ts
```diff
@@ -0,0 +1,211 @@
...
+              af.objectStoreKey
+            )
+          : "";
+      } catch (e) {
+        const errorString = e instanceof Error ? e.message : String(e);
+
+        await this.auditLogService.completeReleaseAuditEvent(
+          this.edgeDbClient,
+          newAuditEventId,
+          8,
```
DoxasticFox, line 96: Is `8` the right setting? `audit-log-service.ts` has this line without explaining what the numbers mean:

https://github.com/umccr/elsa-data/blob/715b1e3b3d9829523a92fd310e0f17c76d47716e/application/backend/src/business/services/audit-log-service.ts#L31

application/backend/src/test-data/insert-test-data-gs.ts
```diff
@@ -0,0 +1,237 @@
...
+                      e.insert(e.consent.ConsentStatementDuo, {
+                        dataUseLimitation: sc,
+                      })
+                    )
+                  ),
+                })
+              : undefined,
+          artifacts: await createArtifacts(
+            createFile(
+              `gs://umccr-10g-data-dev/${specimenId}/${specimenId}.hard-filtered.vcf.gz`,
```
DoxasticFox, line 70: This file is almost entirely copied from https://github.com/umccr/elsa-data/blob/cb3421493ee37032637fa4f07b29e855698085f9/application/backend/src/test-data/insert-test-data-10g.ts but with some values changed (e.g. `s3://...` to `gs://`) such that the objects will be signed using GCP instead of AWS.

application/frontend/src/pages/releases/detail/presigned-url-form.tsx
```diff
@@ -59,22 +59,22 @@ export const AwsS3PresignedForm: React.FC<Props> = ({
             form POSTS can be converted natively into a browser file save dialog
              i.e. if the POST returned a Content-Disposition header */}
       <form
-        action={`/api/releases/${releaseId}/aws-s3-presigned`}
+        action={`/api/releases/${releaseId}/presigned`}
```
DoxasticFox, line 62: I wasn't sure whether it was more appropriate to have two separate panels for each cloud provider. I assumed the underlying provider was a detail the users didn't need to worry about. The correct provider for each presigned URL in the TSV is inferred from the scheme (e.g. `s3://` or `gs://`) in the `objectStoreUrl`. So if a dataset in a release contains URLs of both providers (is that possible?), then the TSV would too. i.e. there's nothing in the presigning logic to say all URLs in a TSV should either be from S3 or GS.

application/backend/src/business/services/gcp-presigned-urls-service.ts
andrewpatto: I'm happy with this - happy even if the subtract time is 1 hour less than 7 days rather than 5 mins (if we are worried about elsa server time - though I think there's a whole bunch of other things that will be failing if the server is way out - not sure the server can make TLS connections to EdgeDb for instance if its clock is wildly skewed).

Is the expiry time allowed related to the lifespan of the (possibly temporary) credentials of the running service? i.e. if I am running in GCloud itself and have a 1 day credential - can it sign out to 7 days? (this is an issue we have to workaround in the AWS signing and just wondering?)

andrewpatto: ignore me - I just read your readme!

application/backend/src/business/services/_release-file-list-helper.ts
andrewpatto: no, s as a protocol would be a mistake.. not sure why the regex had that.. copy/paste mistake probably.

application/backend/src/business/services/_release-file-list-helper.ts
andrewpatto: Nah.. that's fine.. not sure why there was a leading _.. possibly webstorm was complaining about it or something..

application/frontend/src/pages/releases/detail/presigned-url-form.tsx
andrewpatto: Yeah good question.. I think we will need later to add some logic that does a cross match between the object URL schemas and *the credentials we have*.. i.e if asked to sign a gs:// but we only have AWS credentials we will need that to either warn or fail nicely. But I don't object to the 'signing' being merged to be cross platform

application/backend/src/test-data/insert-test-data-gs.ts
```diff
@@ -0,0 +1,237 @@
...
+      externalIdentifiers: makeSystemlessIdentifierArray("GS"),
+      description: "A Dataset to test Google Storage",
+      cases: e.set(
+        await makeCase(
+          "SINGLETONCHARLES",
+          "CHARLES",
+          "male",
+          "HG00096",
+          51808533335,
+          "bfce3cfa4f0265866fae8f7a653cca95-3089", // pragma: allowlist secret
```
andrewpatto, line 109: We possibly should look into GS checksumming.. these are AWS E-tags which get auto made by S3. Not sure if the exact same concept converts over to GS. Maybe we need to GS E-TAG (or equivalent) in our 'checksum types'?

(not important for this PR though)

application/backend/src/test-data/insert-test-data-gs.ts
andrewpatto: Did you actually duplicate the data into a real GS bucket? (and if not, should be make a bucket with real(ish) data in GS?)

DoxasticFox: Yep. [Here's](https://storage.googleapis.com/umccr-10g-data-dev/HG03433/HG03433.bam?X-Goog-Algorithm=GOOG4-RSA-SHA256&X-Goog-Credential=elsa-data-test-acct%40numeric-melody-288107.iam.gserviceaccount.com%2F20230207%2Fauto%2Fstorage%2Fgoog4_request&X-Goog-Date=20230207T042147Z&X-Goog-Expires=604501&X-Goog-SignedHeaders=host&X-Goog-Signature=bfeff4875c1b41866bdf9fbce51024608835365f7307e51bbb625978261aaedae0a59f5d30c15eb363daf845f83df91540e1abe89a33dbbbe15f52999c10c729137cc3dc9d972c9513c0f5aa29a9de885fb3806f6478a4e59a960d72ebb384195c527ca3edd88d44c31f4d0c9d1e0938bcafa76d482884a5033baadf8bcc3c6b4c7efafd8f7eea3cb173892f82c8ead54b41d84c0b5dc8cf29930d8135df69e89e8231d0e049e94928ebb9ca6c13fa25d35ec59c293a9d6163acb8d3260252c204040998a778af87d3a0f019964300dced09d64c3ddce0451898047ea545315d22e8e81728d3a01f8a0a3b1bfbeeaf739641851ec7f9bb3374f58d5990648f76) one of the presigned URLs which Elsa generated.

DoxasticFox: Is it easy for you to test that PR still works on AWS? I just fixed a [bug in my code](https://github.com/umccr/elsa-data/pull/198/commits/ab8766541ed123740dc5d3266f2eebb3121d01fd) which would have affected AWS presigning

application/backend/src/business/services/presigned-urls-service.ts
mmalenic: Although I'm not sure why the outcome codes are those numbers, a description of the numbers is here: https://github.com/umccr/elsa-data/blob/ab8766541ed123740dc5d3266f2eebb3121d01fd/application/backend/dbschema/audit.esdl#L47-L55

I think these are based on: https://www.hl7.org/fhir/valueset-audit-event-outcome.html

In this case I think 8 is the right code because it indicates a failure, although it could potentially be 4 too depending on what the error is.

andrewpatto (approved):

application/backend/src/business/services/presigned-urls-service.ts
andrewpatto: It all originated in an old Health systems (MRI machine interfaces) standard for audit trails.. but its the only standard I know that we can lean on. Obviously we could also just do our own thing if it all becomes too painful.. (but equally we'd just be assigning random integers to mean success, failure etc)

PR #199, opened 2023-02-09
build(backend): upgrade edgedb client library to 1.0

mmalenic:
This PR upgrades the edgedb client library to `^1.0`, which includes support for features like query generators.
### Changes
* The npm `edgetypes` script now generates edgeql-js, queries, and interfcaes.
* Incompatible imports with interfaces.
* `filter_single` when selecting single objects.

See edgedb-js [v1.0.0](https://github.com/edgedb/edgedb-js/releases/tag/v1.0.0) breaking changes.

I figured I would create a separate PR so that everyone could see the changes.


application/backend/src/business/services/australian-genomics/s3-index-import-service.ts
```diff
@@ -409,14 +410,10 @@ export class S3IndexApplicationService {
...
+    if (!pedigreeIdArray.length) {
       pedigreeUUID = null;
-    } else if (Array.isArray(pedigreeIdArray) && !pedigreeIdArray.length) {
-      pedigreeUUID = null;
-    } else if (Array.isArray(pedigreeIdArray)) {
-      pedigreeUUID = pedigreeIdArray[0].id;
     } else {
-      pedigreeUUID = pedigreeIdArray.id;
+      pedigreeUUID = pedigreeIdArray[0].id;
     }
```
mmalenic, line 417: I think this logic is equivalent, because now the cardinality and type can be directly inferred from the query.

andrewpatto:
There will be some Dockerfile commands that also need to be updated to the new query generator..

andrewpatto:
Otherwise looks good - if it compiles and runs then that's further than I got with 1.0.0 of the new JS libraries..

andrewpatto (approved):

DoxasticFox (approved):

DoxasticFox (approved):

application/backend/src/business/services/australian-genomics/s3-index-import-service.ts
williamputraintan: I think it should be `!pedigreeIdArray?.length` as there is a possibility the `pedigreeIdArray` is `null`

williamputraintan: It's probably the best to change the query to `.assert_single()`. I'll make a commit to this PR.

mmalenic: Sure, go for it.

application/backend/tests/service-tests/ag-s3-import.test.ts
```diff
@@ -163,11 +163,11 @@ describe("AWS s3 client", () => {
       size: 0,
       checksums: [
         {
-          type: storage.ChecksumType.MD5,
```
williamputraintan, line 166: Should this be `"MD5"`?

application/backend/tests/service-tests/ag-s3-import.test.ts
```diff
@@ -163,11 +163,11 @@ describe("AWS s3 client", () => {
...
+          type: "AWS_ETAG",
```
williamputraintan: Should this be "MD5"?

mmalenic: Ahh yes, good catch. Thanks!

williamputraintan (approved):

PR #186, opened 2023-01-17
An API for release manifests (htsget)

andrewpatto:
Added activation schema
Wrote manifest producer and test cases

PR #203, opened 2023-02-17
GCP equivalent of AWS S3 Access Point Sharing

DoxasticFox:
This feature lets users share Google Storage objects with anyone who has a gmail account. Demo:

[simplescreenrecorder-2023-02-17_11.19.46.webm](https://user-images.githubusercontent.com/11529449/219518297-17daec8a-19e2-4b6d-93c2-95cc18403045.webm)

This solution sucks because permissions aren't applied transactionally. If someone enters a mixture of IAM principals which do and don't exist, then applying permissions will fail and (almost surely) be partially applied. Automatic rollbacks are impossible in that case because you don't know which ACLs were added by Elsa. The proper solution would have been to create an IAM group for each release, and link users to objects using those groups. Unfortunately, Google requires you to subscribe to Google Identity or Google Workspace to programmatically create and modify groups. Hence, this PR.

Another thing to note is that I haven't hooked up the UI components to generate a TSV manifest in this PR. My initial thought was that I could annotate objects with metadata indicating their presence in a release then search for the annotated objects. Unfortunately, GCP doesn't provide a way to search for objects. We'd need to download a list of objects in the release then query their metadata one-by-one to determine their membership. IMO a much better solution would be to use the extensions to the release service (which started in #186 and keep track of objects in a release) to generate the TSVs. That part of the release service seems like a work in progress, so I'll follow up on TSV generation in a later PR.

### DIFF
--- a/application/backend/src/business/services/gcp-storage-sharing-service.ts
+++ b/application/backend/src/business/services/gcp-storage-sharing-service.ts
@@ -1,0 +1,268 @@
+import { AuthenticatedUser } from "../authenticated-user";
+import * as edgedb from "edgedb";
+import { inject, injectable, singleton } from "tsyringe";
+import { UsersService } from "./users-service";
+import { GcpBaseService } from "./gcp-base-service";
+import { AuditLogService } from "./audit-log-service";
+import { Storage } from "@google-cloud/storage";
+import {
+  getAllFileRecords,
+  ReleaseFileListEntry,
+} from "./_release-file-list-helper";
+import pLimit, { Limit } from "p-limit";
+import { Metadata } from "@google-cloud/storage/build/src/nodejs-common";
+
+@injectable()
+@singleton()
+export class GcpStorageSharingService extends GcpBaseService {
+  storage: Storage;
+  globalLimit: Limit;
+  objectLimits: { [uri: string]: Limit };
+
+  constructor(
+    @inject("Database") protected edgeDbClient: edgedb.Client,
+    private usersService: UsersService,
+    private auditLogService: AuditLogService
+  ) {
+    super();
+
+    this.storage = new Storage();
+
+    // GCP probably rate-limits API requests, otherwise their `Storage` library
+    // wouldn't have retry logic and tests for rate-limit exceptions.
+    // Unfortunately, I can't find the documentation to say what the limits are.
+    // A different GCP feature, OAuth, has a quota which is "Dependent on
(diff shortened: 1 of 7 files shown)


application/backend/src/business/services/gcp-storage-sharing-service.ts
```diff
@@ -0,0 +1,268 @@
...
+
+    // GCP probably rate-limits API requests, otherwise their `Storage` library
+    // wouldn't have retry logic and tests for rate-limit exceptions.
+    // Unfortunately, I can't find the documentation to say what the limits are.
+    // A different GCP feature, OAuth, has a quota which is "Dependent on
+    // application history, developer reputation, and riskiness".
+    // (Source: https://support.google.com/cloud/answer/9028764?hl=en). So there
+    // might not be a concrete answer as to what the relevant limits are.
+    const concurrencyLimit = 10;
+    this.globalLimit = pLimit(concurrencyLimit);
```
andrewpatto, line 39: Nice

andrewpatto:
How long might this application process take if there were say 10,000 files? I presume given the rate limiting - it could take minutes? In which case it might be safer if we move this code to be a background task. But hold off on that for the moment - it won't really change the actual service itself - just where it gets called/invoked from. So happy with this PR to go in.

andrewpatto:
And agree with the use of the manifest generator - the intention is that eventually TSVs etc will all come from that. So happy for that to be deferred.

andrewpatto (approved):

andrewpatto:
I presume we will find out soon enough - but am guessing there is no requirement for google cloud credentials to be present if not actually executing the API stuff? For dev work I mean, some of UMCCR devs may have no gcloud present at all on their systems..

DoxasticFox:
>How long might this application process take if there were say 10,000 files?

Updating permissions on 20 objects generally took a second or two, so I'm guessing about 10 minutes. I raised a ticket (https://github.com/umccr/elsa-data/issues/205).

How many files do releases have in the worst and average cases?

>am guessing there is no requirement for google cloud credentials to be present if not actually executing the API stuff?

I haven't tested on a machine without gcloud, but it should be fine. If Elsa crashes on other devs' machines, I messed up and you should ping me on Slack so I can fix it ASAP

DoxasticFox:
I'll wait for @mmalenic to merge https://github.com/umccr/elsa-data/pull/201 given that I rebased this PR on that

PR #208, opened 2023-02-24
Modernise the docker publish with some newer actions

andrewpatto:
Use git token credentials rather than personal ones

### DIFF
--- a/.github/workflows/docker-image-build.yml
+++ b/.github/workflows/docker-image-build.yml
@@ -50,12 +50,14 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Build and Push
-        uses: docker/build-push-action@v4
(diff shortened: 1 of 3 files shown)


andrewpatto:
Doing some loops here to trigger docker builds

andrewpatto:
Make the docker build params mandatory so we don't ever forget them

andrewpatto:
Use some new actions from Docker that help create tags based off github refs

andrewpatto:
Increase how often we publish - pretty much every commit goes to images tagged with PR name
(may have to dial back but will monitor)

andrewpatto:
I won't be able to test the 'release' build process until this is merged - so we may need another round of changes if that doesn't work

DoxasticFox (approved):

PR #210, opened 2023-02-27
we should start using CSP for added security

andrewpatto:


andrewpatto:
I had disabled CSP in early work - but I think Elsa is now mature enough we can enable - and if we run into issues with anything specific we should deal with them in the CSP policy (i.e. adding others sources etc)

williamputraintan (approved):

PR #217, opened 2023-03-02
feat: add separate audit event page for non-release related events

### LINKED ISSUE #40: Add a new audit log for storing audit events that are *not* part of a release
We need a mechanism similar to the 'release' audit mechanism - but for more general purpose events that occur *outside* any release context.

- [ ] login
- [ ] logout
- [ ] change user permissions
- [ ] admin level activities like load new dataset

mmalenic:
Closes #40 

This PR adds a new audit event page which can view audit events that are not just part of a release.

## Changes
* Adds a new top level route for general audit events - release audit events in the releases page remain unchanged.
* New route contains an audit table which can view audit events associated with a user, but not related to a release (e.g. login or logout events)
* A filter menu can also be used to view release events that are related to a user.
* Super admins can view all events of any user, and system level events (e.g. email sent event). 
    
### Backend
* A new `UserAuditEvent` database type representing audit events that are associated with a `User`.
* A `SystemAuditEvent` type which has no associated user.
* Releases have a back link to participating users to help with querying all release events related to a user.
* Audit events are added for:
    * Login.
    * Logout.
    * Sync database.
    * Add database.
    * Delete database.
    * User added to release.
    * Change user permission (although this has no uses yet).
    * Send email.
 * Add a new `ALLOWED_VIEW_AUDIT_EVENTS` allow constant, which allows viewing audit events from any user. Currently the `Super admin` can view all audit events, and other users can only view events relating to themselves.
    
### Frontend
* The audit event table is reused in the releases page and the new audit events page. 
* The filter menu allows filtering `release` or `user` events, and `system` or `all` events if the user is a super admin.
* The `all` filter enables viewing any audit event, however the other filters still need to be checked to view a specific type of event. Maybe this should be clearer in the UI?
* View for a regular user:
image
* View for a user with `ALLOWED_VIEW_AUDIT_EVENTS`, e.g. a super admin:
image

## Improvements/Considerations
* The filtering is all or nothing, and there is no way to view audit events for a specific group of users.
* Searching the table, or using custom filters is not implemented - this could be used filter/find specific audit events.
* Small UI improvements could be made, such as adding loading wheels when applying a new filter, adding a refresh button, or a download data button.


application/backend/dbschema/queries/insertReleaseAuditEvent.edgeql
```diff
@@ -0,0 +1,11 @@
+# Insert a release audit event.
```
andrewpatto: Nice - I've just started adding .edgeql for the user management queries - I like the approach - better than the Typescript builder

andrewpatto: We possibly need a decision on how to name query files - and whether they should live in a folder structure etc..
(i.e do we want all the 'insert' queries grouped together or would we prefer all the 'user' queries grouped together)

application/backend/dbschema/audit.esdl
```diff
@@ -6,12 +6,10 @@ module audit {
 
     abstract type AuditEvent {
         # who initiated the action being audited
-
-        required property whoId -> str;
-        required property whoDisplayName -> str;
+        property whoId -> str;
```
andrewpatto: Happy to be overruled - but initial thought here was whether this is the right modelling..

Because when we *do* have an audit event performed by a user - these *are* absolutely required (whereas this would allow mistakes like whoId filled in, but whoDisplayName not.

We could introduce a new audit level one level above the base (PerformedAuditEvent?? - better name needed)

AuditEvent -> PerformedAuditEvent (with whoId required) -> DataAccessAuditEvent...
AuditEvent -> SystemAuditEvent

(please come up with a better name for the event though!)
(or happy to defer to revisit in a different PR - though we need to nail this down before going to MVP)

application/backend/src/api/api-auth-routes.ts
```diff
@@ -271,10 +302,14 @@ export const callbackRoutes = async (
 
     if (isa) {
       allowed.add(ALLOWED_CHANGE_ADMINS);
+      allowed.add(ALLOWED_VIEW_AUDIT_EVENTS);
```
andrewpatto, line 305: Nice - we obviously need to flesh this out as we explore the user roles etc - but this is a good start!

application/frontend/src/index-router.tsx
```diff
@@ -127,7 +128,7 @@ export function createRouter(addBypassLoginPage: boolean) {
           <Route path={`releases`}>
             <Route index element={<ReleasesPage />} />
             <Route
-              path={`:releaseId`}
```
andrewpatto: ??

application/backend/src/business/services/australian-genomics/redcap-import-application-service.ts
```diff
@@ -324,6 +323,17 @@ ${roleTable.join("\n")}
...
+                userAuditEvent: {
+                  "+=": e.insert(e.audit.UserAuditEvent, {
+                    whoId: dbUser.subjectId,
+                    whoDisplayName: dbUser.displayName,
+                    occurredDateTime: new Date(),
+                    actionCategory: "E",
+                    actionDescription: "Add user to release",
+                    outcome: 0,
+                    details: e.json({ role: role }),
+                  }),
```
andrewpatto: I worry that the repetition of this pattern might be hard to maintain (well that it might drift between the different services that have this code) - not that I have a better solution

andrewpatto:
Are the general view of 'release events' useful without a column that says what release it happened on?
(i.e. "Viewed Release" doesn't help if I don't see the id of the release viewed)

andrewpatto (approved):

andrewpatto:
Excellent progress on the auditing - we may need to loop back around for another PR down the track but happy to let this settle for a bit

andrewpatto:
We should get a handle at some point on the behaviour of deletion and user audit events. i.e. once I have a user audit event - can I delete the corresponding user? Technically what currently happens - and what is the desired behaviour (I'm not sure I know how I want it to behave?)

application/backend/src/test-data/insert-test-data.ts
```diff
@@ -22,6 +22,8 @@ import {
   TEST_SUBJECT_3_DISPLAY,
   TEST_SUBJECT_3_EMAIL,
 } from "./insert-test-users";
+// @ts-ignore
```
andrewpatto: ??

application/backend/dbschema/queries/insertReleaseAuditEvent.edgeql
mmalenic: We could copy the structure of the typescript client library [queries](https://github.com/umccr/elsa-data/tree/feat/audit-event-page/application/backend/src/business/db), e.g. group in folders by user, release, etc.

application/backend/dbschema/audit.esdl
mmalenic: Yes, I think this makes sense. I initally had an [`OwnedAuditEvent`](https://github.com/umccr/elsa-data/blob/e4e31009d5d8984baa4abd8fdb9b15dc29d36c12/application/backend/dbschema/audit.esdl#L57-L62) and `AuditEvent` structure, however it complicated writing out the queries a bit, so I went with this approach. However, I think you're right, having that structure is closer to the correct modelling I think.

application/backend/src/business/services/australian-genomics/redcap-import-application-service.ts
mmalenic: Yeah that's a good point, maybe a function to create to create this event somewhere would help.

application/backend/src/test-data/insert-test-data.ts
mmalenic: Typescript currently complains if there is a `.ts` at the end of the import, and without it, there is an error trying to import a `.edgeql` file. So the solution was to include the `.ts` so that it could be properly resolved to the correct file, while silencing typescript until it [supports](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0-beta/#allowimportingtsextensions) a `.ts` at the end of an import.

williamputraintan (approved):

application/backend/src/test-data/insert-test-data.ts
andrewpatto: Ah ok - I think getting Edge to generate just a single queries.ts file solves this (and is probably neater).. so everything just imports from "../../dbschema/queries".
This is how I set it up in my new PR - so leave this PR as is and I'll change it over when I put mine through..

mmalenic: Okay. If we have a lot of queries, wouldn't the single queries file get a big long though?

DoxasticFox (approved):

application/backend/dbschema/queries/insertReleaseAuditEvent.edgeql
andrewpatto: I mean even more than that - like literally what is the naming conventions of the files..
e.g. insertReleaseAuditEvent v releaseAuditEventInsert!
especially where we start to get more complex queries
getAllReleaseAuditEventsWithPaging

I have no answer - I'm even inconsistent with my own naming

application/backend/dbschema/audit.esdl
andrewpatto: I like OwnedAuditEvent as a name.

application/backend/src/test-data/insert-test-data.ts
andrewpatto: Meh.. its autogenerated so its not like we have to maintain its content by hand.. i.e. noone needs to ever view it

andrewpatto: Its make the docker build slightly more straightforward as we don't have to tease out all the many generated TS files - just bring one file over

application/frontend/src/index-router.tsx
mmalenic: I initially had this so that `AuditEventTable` could `useParams` in a generic context. I.e. it didn't matter if the path contained a `releaseId` or a `userId`. But I moved the logic around anyway so this is not needed.

application/backend/src/test-data/insert-test-data.ts
mmalenic: Ahh true. I misunderstood and thought that the EdgeQL queries had to all be in one file, but that's not the case.

application/frontend/src/index-router.tsx
andrewpatto: All good - am redoing some of the release routing anyhow - so don't worry about finessing this

application/backend/dbschema/audit.esdl
mmalenic: I'm having issues actually constructing queries involving this kind of hierarchy. The problem is that it is not possible to create a union of `OwnedAuditEvent` and `AuditEvent` - which is required for the way that `AuditEventTable` shows and filters results. I also don't think it is possible to create a union of only `AuditEvent`, where `whoId` and `whoDisplayName` are computed properties. 

I know this isn't ideal, but would it be okay for there to only be one inheritance level (e.g. `AuditEvent`), where `whoId` and `whoDisplayName` are required, but `SystemAuditEvent` sets these to an empty string? This is more or less the way the events are displayed in the table anyway.

mmalenic:
I've added a few functions for common audit events, and included more information about the dataset and release in the `UserAuditEvent`. At some point, I think it would be good to include a link from the `AuditEventTable` entry for a `UserAuditEvent` to the actual release or dataset.

andrewpatto (approved):

application/backend/dbschema/audit.esdl
andrewpatto: I mean the EdgeDb queries don't need to match up directly to a frontend view of the world..

that is - there is another mapping layer (API/service) in between the db queries and frontend. Surely any query anomalies can be dealt with there?

andrewpatto: I mean the API call for AuditEventTable - could literal execute two different queries and hand merge in the service/API..

but I'm pretty sure it can be done with one query - do you want to put a discussion in Slack of what the query currently looks like and maybe we can workshop it?

mmalenic: Sure, I think it becomes difficult because the union needs to be sorted/filtered on the whole set rather than individually.

PR #220, opened 2023-03-06
`releaseIdentifier` vs `releaseUUID`

### LINKED ISSUE #212: Convert to using our Release Identifier rather than edge id in all routes/api calls to do with releases
We are currently using edge db release id in our react routes and API calls.

However - I think we want to assign a user visible release identifier (e.g "AGRTE3AF") to each release and have that be the main visible identifier as far as the user is concerned. In which cases - I think it would benefit to replace it entirely from front end to back.

I mean we have a getReleaseInfo in the backend - we simply just get it to search by identifier not id - so I don't think this is a massive change.

williamputraintan:
Migrate releaseIdentifier from UUID to `RXXX` (where XXX is n<sup>th</sup> of the release record) .

The R stands for Release but is customisable from the config file.

Fix #212

### DIFF
--- a/application/backend/src/business/db/release-queries.ts
+++ b/application/backend/src/business/db/release-queries.ts
@@ -42,11 +42,11 @@ export const allReleasesSummaryByUserQuery = e.params(
  */
 export const touchRelease = e.params(
   {
-    releaseId: e.uuid,
+    releaseId: e.str,
   },
   (params) =>
     e.update(e.release.Release, (r) => ({
-      filter: e.op(r.id, "=", params.releaseId),
+      filter: e.op(r.releaseIdentifier, "=", params.releaseId),
       set: { lastUpdated: e.datetime_current() },
     }))
 );
--- a/application/backend/src/business/db/release-queries.ts
+++ b/application/backend/src/business/db/release-queries.ts
@@ -50,3 +50,25 @@ export const touchRelease = e.params(
       set: { lastUpdated: e.datetime_current() },
     }))
 );
+
+/**
+ * Find the next release identifier based on release count.
+ * @param prefix The prefix id before the release count.
+ * @returns
+ */
+export const getNextReleaseId = (prefix: string = "R") => {
+  const minDigitLength = 3;
+
+  const nextIdString = e.to_str(e.op(e.count(e.release.Release), "+", 1));
+
+  // Formatting string to add leading 0s if len(nextId) < minDigitLength
+  const formatIdString = e.op(
+    e.str_pad_start(nextIdString, minDigitLength, "0"),
+    "if",
+    e.op(e.len(nextIdString), "<=", minDigitLength),
+    "else",
+    nextIdString
+  );
+
+  return e.op(e.str(prefix), "++", formatIdString);
+};
--- a/application/backend/src/business/db/release-queries.ts
+++ b/application/backend/src/business/db/release-queries.ts
@@ -27,7 +27,7 @@ export const allReleasesSummaryByUserQuery = e.params(
         // as release identifier is essentially a meaningless random string
         order_by: {
           expression: rp.releaseIdentifier,
-          direction: e.ASC,
+          direction: e.DESC,
         },
         limit: params.limit,
         offset: params.offset,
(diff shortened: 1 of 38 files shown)


andrewpatto:
I actually thought we might only do this at the boundaries of the system - so when it gets to an API - that uses the release identifier - the first operation would be to map it to a edgedb id (and so all the queries would stay the same).

That said - what you've done will work - are there any optimisations that EdgeDb will be able to use if it was using its native id throughout?

application/backend/src/business/db/release-queries.ts
```diff
@@ -42,11 +42,33 @@ export const allReleasesSummaryByUserQuery = e.params(
...
+
+/**
+ * Find the next release identifier based on release count.
+ * @param prefix The prefix id before the release count.
+ * @returns
+ */
+export const getNextReleaseId = (prefix: string = "R") => {
+  const minDigitLength = 3;
+
+  const nextIdString = e.to_str(e.op(e.count(e.release.Release), "+", 1));
```
andrewpatto, line 62: edgedb has an autoincrementing counter type I believe (though I was unsure how to get it to work - but I think the docs might be better now)

andrewpatto: (the danger here is that two simultaneous transactions can technically generated the same 'nextId')

andrewpatto (approved):

andrewpatto:
I mean the PR is fine - I wonder if there is some sort of naming so we can make it super clear that `releaseId` *isn't* referring to the `.id` of the Release

williamputraintan:
> I actually thought we might only do this at the boundaries of the system - so when it gets to an API - that uses the release identifier - the first operation would be to map it to a edgedb id (and so all the queries would stay the same).

I thought that, but I think it is more value having a readable releaseId in audits and also on the cloudFormation stack? I just thought cloudFormation stack need to truncate names if it is too long.

application/backend/src/business/db/release-queries.ts
williamputraintan: Yup, at least the releaseIdentifier schema is unique so at least it will fail.

andrewpatto:
yep - good point! consistency throughout is a good goal. We will just have to be very careful with the `.id` the comes back in queries to do with releases (as in whenever you a selecting from Release - EdgeDb by default is giving back sets of `.id`

application/backend/src/business/db/release-queries.ts
andrewpatto: Yep - no long term problem because of the uniqueness constraint - but if you can get the native counter generator working that would be best practice

andrewpatto:
Given the number of changes in the PR - happy if you want to get this merged - and then do a smaller patch PR for fixing the 'counter' generator

application/backend/src/business/db/release-queries.ts
williamputraintan: > edgedb has an autoincrementing counter type I believe (though I was unsure how to get it to work - but I think the docs might be better now)

I suppose referring to [this](https://www.edgedb.com/docs/stdlib/sequence#type::std::sequence)? I am also looking for examples on how to use this.

williamputraintan:
cc: @mmalenic @DoxasticFox 

Just a side note, from this PR:
- `releaseId` is now referring to the `releaseIdentifier` property from the release schema
- `releaseUuid` is the one referring to `.id` from the schema.

williamputraintan:
> I wonder if there is some sort of naming so we can make it super clear that releaseId isn't referring to the .id of the Release

How about tag (releaseIdentifier -> releaseTag)? Just looking out for synonym of Id and tag might do (or perhaps it sound odd)?

andrewpatto:
Yeah something like that.

releaseTag, releaseKey, releaseHandle, releaseLabel

Tag is not bad - but tags don't imply the uniqeness I don't think (I expect to have multiple tags for an entity)

application/backend/src/business/db/release-queries.ts
williamputraintan: Can't figure this out in TS. Followed on the NOTES here and translate to typescript

```
  const releaseCounterScalar = e.select(e.schema.ScalarType, (st) => ({
    id: true,
    filter: e.op(st.name, "=", "release::ReleaseCounterSequence"),
  }));
  const nextIdString = e.to_str(e.sequence_next(releaseCounterScalar));

```

But it still complain on non-assignable types. Will revisit this again.

Also to note, the sequence here is individual than the release records meaning `e.delete(...` won't reset the counter but would need to use the `sequence_reset()` function.

PR #222, opened 2023-03-08
Releases dashboard page needs proper paging

andrewpatto:
Change the base query for getting all releases - and thereby introduce proper release paging.


andrewpatto:
Also brings through the 'lastUpdated' field of releases (now that we are trying to keep that correct) and we use that field for sorting releases.

DoxasticFox (approved):

williamputraintan (approved):

PR #234, opened 2023-03-10
Move Cohort Construction / AccessPoint actions to TRPC

andrewpatto:
There are some notable actions that are not at all RESTful and we can move to TRPC


application/frontend/src/pages/releases/bulk-selector-sub-page/bulk-box/bulk-box.tsx
```diff
@@ -36,15 +34,17 @@ export const BulkBox: React.FC<Props> = ({ releaseKey, releaseData }) => {
     setError({ error: null, isSuccess: true });
   };
 
-  const applyAllMutate = useMutation(
-    axiosPostNullMutationFn(`/api/releases/${releaseKey}/jobs/select`)
```
andrewpatto, line 40: One of the beauties here is that we do not need to keep these arbitraty URLs synchronised between the backend and frontend

DoxasticFox (approved):

williamputraintan (approved):

PR #235, opened 2023-03-10
Wrapped activation/deactivation in audit events

andrewpatto:


andrewpatto:
I am not particularly happy with the audit try/catch/success/fail pattern (its a bit verbose and repetitive) but its what we currently have..

DoxasticFox (approved):

application/backend/src/audit-helpers.ts
```diff
@@ -1,25 +1,93 @@
...
+  auditLogService: AuditLogService,
+  edgeDbClient: Client,
+  user: AuthenticatedUser,
+  releaseKey: string,
+  actionCategory: audit.ActionType,
+  actionDescription: string,
+  initFunc: () => Promise<T>,
+  transFunc: (tx: Transaction, a: T) => Promise<U>,
+  finishFunc: (a: U) => Promise<V>
+) {
```
mmalenic: You could potentially extract this into a Class, as part of `AuditLogService`? Which might reduce some of the parameters. Similar with the other functions in `audit-helpers`.

Not sure if this makes `AuditLogServive` too large then, maybe there should be another Class to break it up a bit.

mmalenic (approved):

application/backend/src/audit-helpers.ts
andrewpatto: Yeah I moved some of the callers of this into AuditLogService - but just left this block here for the moment.. but agree - probably should move it into AuditLogService.. and the other functions in audit helps will go once everything uses this pattern..

PR #239, opened 2023-03-12
bug(frontend): audit sub page showing too many audit event

### LINKED ISSUE #236: Audit log sub page (part of releases) needs small fixes
I'm sure I broke it when making the sub page (possibly a poor merge on my part) - but I think the audit log sub page that you can see as part of a release is showing more events than just the release (I can see maybe 4 "created release" events which doesn't seem right).

Just needs a quick check of the settings passed into the AuditEventTable object I think to restrict it to just activities of the current release.

mmalenic:
Closes #236 

Changes the audit event table to use the releases api path when it is rendered in the releases page. This fixes the audit table fetching incorrect events.


andrewpatto (approved):

mmalenic:
#### Update: small additional UI changes
* Render audit log details as part of the `Route` hierarchy it belongs to.
* Move audit log filter menu to the right side.

PR #242, opened 2023-03-13
Make pino transport inherit global log level

### LINKED ISSUE #229: DEBUG Level logging doesn't seem to work even though that is the logger level
In the setup of the pino logger in entrypoint - we definitely seem to be passing in a log level of "debug" (in dev).

However, debug level logs don't seem to be displayed (see the log level tests that happen immediately after we set up the logger)

DoxasticFox:
Closes https://github.com/umccr/elsa-data/issues/229.

Pino transport targets have their own log levels which are independent of the global `level:` setting. This PR causes the transport targets to inherit the global level unless those targets set their own level.

### DIFF
--- a/application/backend/src/bootstrap-settings.ts
+++ b/application/backend/src/bootstrap-settings.ts
@@ -65,6 +65,11 @@ export async function bootstrapSettings(config: any): Promise<ElsaSettings> {
       },
     ];
 
+  const logLevel = config.get("logger.level");
+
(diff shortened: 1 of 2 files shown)


andrewpatto:
I suspected it might be something like this but pored through all the documentation and couldn't see anything calling this out..

andrewpatto (approved):

Issue #37, opened 2022-08-31
A general user administration page that allows suitable people to change other permissions/roles


andrewpatto:
The super admins need at a minimum the ability to bless other users with some forms of admin rights

andrewpatto:
There is a base UI for this - but is dependent on decisions re: user/roles to actually complete

PR #245, opened 2023-03-15
Update Role/Permissions

williamputraintan:
- Updated user's permissions to have (superAdmin have the following all `true`):
	- isAllowedSyncDataAccessEvents
	- ~isAllowedImportDataset~ isAllowedRefreshDatasetIndex
	- isAllowedCreateRelease
    - isAllowedViewAllAuditEvents
    - isAllowedViewDatasetContent
    - isAllowedViewUserManagement
    - isAllowedViewAllReleases
 - Added auth permission checks on service layer
 - Tailored some service to take into account the permission above (e.g. A user is granted to view all release, therefore, getAllReleases query will return all releases despite not a participants. )
 - Rename role in release:
 	- `DataOwner` -> `Administrator` - Creator of release, promote a `Manager`, finalise the content of releases
 	- `PI` -> `Manager` - view release content, invite people to become a member
 	- Member (stays the same) - View content of the release
 	
 	Details of actions allowed by each role is on [Miro board](https://umccr.slack.com/archives/C03F7BHNJ3X/p1678058784012049) (access link through slack).
 	

Next step:
- The ability to change the permissions above as a superAdmin #37


application/backend/dbschema/queries/release/releaseGetAllByUser.edgeql
```diff
@@ -22,7 +31,7 @@ select {
               datasetUris,
               applicationDacTitle,
               applicationDacIdentifier,
-              role := (select assert_single(.userRole@role)),
+              role := (select assert_single(.userRole@role)) ?? "Viewer",
```
williamputraintan: returning a `Viewer` role if its not in the participant list but allowed to view the release.

application/backend/dbschema/permission.esdl
```diff
@@ -51,24 +51,44 @@ module permission {
...
+
+        required property isAllowedImportDataset -> bool {
            default := false;
         };
-        required property allowedChangeReleaseDataOwner -> bool {
+
+        required property isAllowedCreateRelease -> bool {
             default := false;
         };
+
```
andrewpatto, line 69: The maintenance and evolution of these may be an issue (as in adding features/changing features and permissions becomes something that requires a database migration).. just flagging this - happy to play wait and see at the moment though

application/backend/dbschema/queries/release/releaseGetAllByUser.edgeql
andrewpatto: are the Role enum types in the backend going to be a pain given they won't have "Viewer" listed?
(i.e. whatever the reflected database type for User::role is?)

application/backend/src/business/authenticated-user.ts
```diff
@@ -13,9 +13,13 @@ export class AuthenticatedUser {
       isNil(this.dbUser.displayName) ||
       isNil(this.dbUser.email) ||
       isNil(this.dbUser.lastLoginDateTime) ||
-      isNil(this.dbUser.allowedCreateRelease) ||
-      isNil(this.dbUser.allowedChangeReleaseDataOwner) ||
-      isNil(this.dbUser.allowedImportDataset)
+      isNil(this.dbUser.isAllowedSyncDataAccessEvents) ||
```
andrewpatto, line 16: as per our discussion in slack - it might turn out better that we never actually put these into the Cookie Authenticated user - and we make people always fetch them live from the db on each query?
(happy to leave this in - just wondering about which path we might eventually take?)

application/backend/src/business/exceptions/ag.ts
```diff
@@ -0,0 +1,13 @@
+import { Base7807Error } from "@umccr/elsa-types/error-types";
+
+export class NotAuthorisedImportAGDatasets extends Base7807Error {
```
andrewpatto: Are these exceptions exclusive to AG? I would have thought there is a much more generic concept of
Importing or Refreshing or Updating a dataset (whether it be AG or not). I would have though the exception is general (similarly - I would think the permission is general - I doubt some admins would be able to import AG datasets but not be able to import Garvan datasets - for instance)..

andrewpatto: Question if Import (all through this dataset permissions) is the right word - given the pre-registration of datasets in the config files.. it feels like "import" is an activity that causes the base record to be created.. maybe Refresh? Update? Or something that implies we are importing the index/content? UpdateIndex? RefreshContent?

application/backend/dbschema/queries/release/releaseGetAllByUser.edgeql
williamputraintan: I just had a 2nd thought at least it gives some context of why it is returning these values even though user is not part of the release. But yes could move it at the application layer and just reflect based on db.

application/backend/src/business/authenticated-user.ts
williamputraintan: Yes, I guess could be done on seperate ticket, I was reusing the existence of authenticated User. But possibly yes it makes more sense if hits the API to get latest data.(I think It feels weird if user need to re-login just to get these updated?)

application/backend/src/business/exceptions/ag.ts
williamputraintan: Yes, I just thought of specific exception to be clear of what and why it cannot perform the task. But thinking it now, it probably should be a generic `notAuthroised` to update the index there. (Will change this)

Yup, got it, say maybe `RefreshDatasets` or `RefreshDatasetsIndex`?

application/backend/dbschema/permission.esdl
mmalenic: Here's a question with a similar situation, with some possible alternative [solutions](https://dba.stackexchange.com/questions/62615/how-to-implement-business-logic-permissions-in-postgresql-or-sql-in-general).

I think the top answer suggests using a built-in database solution directly. I know the super admin concept is a config-level construct, but this could potentially also be done with something like [access policies](https://www.edgedb.com/docs/datamodel/access_policies).

andrewpatto (approved):

application/backend/src/business/exceptions/ag.ts
andrewpatto: How about `RefreshDatasetIndex` as the API level operation name

DoxasticFox (approved):

mmalenic (approved):

application/backend/dbschema/permission.esdl
williamputraintan: Yes the `access policies` of edgedb seems nice. Will try this out!

PR #246, opened 2023-03-15
Feature/manifest creation is more generic

andrewpatto:


andrewpatto:
So this is not particularly any new code - it is a rewriting of some existing Typescript queries into a more fully featured Edgeql.

And it introduces a formal type called the `ManifestMasterType` that is a JSON structure holding exactly what has been released when a release is activated.

The idea is that all other manifests must be a transform purely from this master JSON.

Other manifests will be
- htsget manifest
- presigned URL manifest TSV/CSV dump
- identifier manifest (no file access, just data about the identifiers in a tree)

andrewpatto:
We currently have a few other existings helpers (file list helper etc) that should all be deprecated and converted so that everything that releases file all comes from the same source (this helps minimise mistakes in the release logic - the worst mistake being releasing data that shouldn't be released)

williamputraintan (approved):

mmalenic (approved):

DoxasticFox (approved):

Issue #29, opened 2022-08-16
Create a file sharing / auth mechanism for CloudFlare R2


andrewpatto:
Cloudflare R2 promises no egress fees on data (worldwide) - would be good to think of a way that controlled datasets could be shared.

There is some sort of 'lambda' that runs on each request - and can enforce custom auth - but would have to work out how to do passports/visas processing?

andrewpatto:
I mean can we literally make an access key/credentials (however CloudFlare does it) - and just do Object store presigned URLs in the first instance?

andrewpatto:
The access key object store signing will work fine for now so we should enable that (it is already partially written)

PR #247, opened 2023-03-15
Feature/cloudflare keys begin

andrewpatto:
Adds a spot in the config files for R2 signing keys - though no actual implementation yet.

This needs to be in place first so we can edit the shared AWS secrets - and not break other dev instances when they encoutner config fields they don't understand.


andrewpatto:
Also a minor change to move the aws* config settings into a nested structure.

andrewpatto:
Progresses #29

andrewpatto:
Ok this seems to be the flaky E2E test so I will merge once approved

williamputraintan (approved):

mmalenic (approved):

DoxasticFox (approved):

PR #248, opened 2023-03-16
Pretty much allow any form src

andrewpatto:


andrewpatto:
Apparently the behaviour when forms get redirected is inconsistent between browsers - but OIDC works by literally redirecting forms until it gets to the institute login. So this seems to work fine locally but then is failing when deployed.
Am trying here with a very broad approval of anything that is https

andrewpatto:
This has a compile error from other merges - I'll put this one through and fix the compile problem elsewhere

williamputraintan (approved):

mmalenic (approved):

PR #252, opened 2023-03-20
Introduced a delete service mirroring create/install

andrewpatto:
Some service splitting as well


application/backend/jobs/entrypoint-job-handler.ts
```diff
@@ -106,11 +114,11 @@ bootstrapDependencyInjection();
 
             case "CloudFormationInstallJob":
               jobPromises.push(
-                jobsService
```
andrewpatto, line 109: I notice I have managed to call this operation both "install" and "create". Will unify in a future PR.

application/backend/src/business/services/jobs/job-cloud-formation-delete-service.ts
```diff
@@ -0,0 +1,259 @@
...
+    if (theStack.StackStatus === "DELETE_COMPLETE") {
+      return 0;
+    }
+    if (theStack.StackStatus === "DELETE_FAILED") {
+      return 0;
+    }
+
+    console.log(theStack.StackStatus);
+
+    return 0;
```
DoxasticFox, line 209: It's not important because it's just a matter of style, but do the return values work like exit codes in shell scripts. If yes, that's quite "Unix hacker" of you, but it's not very idiomatic TypeScript.

DoxasticFox (approved):
Looks good

application/backend/src/business/services/jobs/job-cloud-formation-delete-service.ts
andrewpatto: Excellent question - I didn't have a good direction for job results so may have reverted to old skool unix. But yes it could do with a refactoring. Will make a ticket for that. Thanks

PR #254, opened 2023-03-20
User Permissions Management

### LINKED ISSUE #37: A general user administration page that allows suitable people to change other permissions/roles

williamputraintan:
- Simplifying permissions
	Last permission's PR introduce a few of `isAllowed...` but I think we could simplify that at least for the first cut.
	- `isAllowedChangeUserPermission` - Only superAdmin have this.
	- `isAllowedRefreshDatasetIndex` - This will include sync access data event from external resource
	- `isAllowedCreateRelease`
	- ~`isAllowedElsaAdminView`~ `isAllowedOverallAdministratorView` - View everything which include user's management, All (releases, datasets, audits)
	I think this is better for MVP minimising the permissions? Thoughts?
	
- Use TRPC for changeUserPermissions action
- A few FE components for AdminView and SuperAdmin to change user's permission

ScreenShots:

Screenshot 2023-03-20 at 19 01 29
Screenshot 2023-03-20 at 19 01 54
Screenshot 2023-03-20 at 19 02 47


application/backend/src/api/api-auth-routes.ts
```diff
@@ -156,10 +157,17 @@ export const apiAuthRoutes = async (
...
     const addTestUserRoute = (
       path: string,
-      authUser: AuthenticatedUser,
-      allowed: string[]
+      subjectId: string,
+      name: string,
+      email: string
     ) => {
       fastify.post(path, async (request, reply) => {
+        const authUser = await userService.upsertUserForLogin(
```
williamputraintan, line 165: Moving SubjectTest upsert here so login could reflect permissions from Db.

application/frontend/src/pages/users-dashboard/other-users/other-users.tsx
```diff
@@ -13,13 +16,48 @@ import {
...
+import { PermissionDialog } from "./permission-dialog";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import {
+  faArrowsRotate,
+  faFolderPlus,
+  faUsersGear,
+  faUsersViewfinder,
+} from "@fortawesome/free-solid-svg-icons";
+
+const permissionIconProperties: {
```
williamputraintan, line 28: Thinking to show permissions from the table view. Having some checkbox seems taking too much space, so decided to use icon instead. It has a tooltip to describe what it represent.

I guess if we have more permissions we could potentially running out of icons.

Thoughts?

Screenshot 2023-03-20 at 19 06 54

andrewpatto: Icons with tooltips looks good for me!

andrewpatto:
the UI looks fantastic

andrewpatto:
I wouldn't have the word `Elsa` in `isAllowedElsaAdminView` - seems like all of the permissions are related 'to Elsa' so not sure why it would be the only one with the name in it.
I think it is fine to merge a lot of the permissions into this for MVP.
`isAllowedOverallAdministratorView`?

application/backend/src/api/api-auth-routes.ts
```diff
@@ -308,15 +340,27 @@ export const callbackRoutes = async (
     // that is - they are a deployment instance level setting
     const isa = isSuperAdmin(settings, authUser);
 
-    if (isa) {
-      allowed.add(ALLOWED_CHANGE_ADMINS);
-      allowed.add(ALLOWED_VIEW_AUDIT_EVENTS);
+    if (authUser.isAllowedCreateRelease || isa) {
```
andrewpatto: I'm not convinced that Super Admins should automatically get these rights.. I think I'd make the argument that being super admin gives you exactly one extra right - the ability to alter other users permissions.. (it should be possible for me to be a super admin but otherwise not see everything)

andrewpatto:
Where is the ok/cancel button on the dialog?

andrewpatto:
I think somewhere we need to emphasise that super admin right comes from a config setting and is not otherwise editable from within the app

application/backend/src/api/api-auth-routes.ts
williamputraintan: Right, I would assume a super admin role could do all things, but I guess you are right. The cookie should still reflect what is in the Db.

I suppose they could promote themself in the `others users` box?

application/frontend/src/pages/users-dashboard/other-users/permission-dialog.tsx
```diff
@@ -0,0 +1,282 @@
...
+        <FontAwesomeIcon icon={faUserGear} />
+      </button>
+      <ErrorBoundary styling={"bg-red-100"}>
+        <SelectDialogBase
+          showing={isDialogOpen}
+          cancelShowing={cancelButton}
+          title="User Configuration"
+          buttons={
+            <>
+              {!isInputDisable && (
```
williamputraintan, line 152: > Where is the ok/cancel button on the dialog?

Had that to appear when editing mode is on. 

Screenshot 2023-03-21 at 09 52 38

williamputraintan: > I think somewhere we need to emphasise that super admin right comes from a config setting and is not otherwise editable from within the app

Yup, I made the `changeOtherUserPermission` mentioned that can only be changed by an "app administrator" I guess could rephrase that to "can only be changed within the app configuration".

I'm assuming for now is `superAdmin` 1:1 to `allowedChangeUserPermission`, as this is the only permission they had exclusively as a superAdmin

application/frontend/src/pages/users-dashboard/other-users/permission-dialog.tsx
```diff
@@ -0,0 +1,282 @@
...
+        setIsSuccess(true);
+        setError({ error: null, isSuccess: true });
+      },
+    }
+  );
+  const onSave = () => {
+    setIsSuccess(false);
+    setIsEditingMode(false);
+    changeUserPermissionMutate.mutate({ userEmail: user.email, ...input });
+  };
```
DoxasticFox: Is there a reason why we haven't used [useCallback](https://react.dev/reference/react/useCallback#usecallback) much in Elsa? I'm asking more generally---Not just referring to this PR

DoxasticFox (approved):
Obviously you and Patto know best how the design should be at a high level, but the code LGTM. +1 about the polish on the UI.

application/frontend/src/pages/users-dashboard/other-users/permission-dialog.tsx
williamputraintan: Hmmmm, I thought the use of `useCallback` is effective when the function is passed to a child component? So I suppose having it used within the component does not affect if it is wrapped under the `useCallback`?

DoxasticFox: >I thought the use of `useCallback` is effective when the function is passed to a child component?

Yeah, and `onSave` _is_ being passed to a child component (i.e. `<button>`) in this case, right?

>So I suppose having it used within the component does not affect if it is wrapped under the `useCallback`?

One of the main uses of `useCallback` is to prevent needlessly re-rendering components. [React's docs](https://react.dev/reference/react/useCallback#skipping-re-rendering-of-components) have a decent explanation of why a re-render might happen if `useCallback` isn't used.

I'm mostly suggesting it as a performance optimisation (which won't affect the functionality of Elsa from the user's perspective). If a needlessly re-rendered component does a network operation in `useEffect` (for example), that's some load on the server which can be avoided. I vaguely remember looking at the "Network" tab in Chrome's dev tools when I was working on previous tickets and noticing some extra API hits, but Elsa's performance seemed good enough, so I just kind of shrugged it off lol

Anyway, avoiding those extra re-renders won't be as easy as putting `useCallback` everywhere and, for my taste, it's uncomfortably easy to break the code by not passing the correct dependencies to `useCallback`. So is it worth it? :man_shrugging:

application/frontend/src/pages/users-dashboard/personal-details-box/personal-details-box.tsx
```diff
@@ -0,0 +1,77 @@
...
+    case ALLOWED_CREATE_NEW_RELEASE:
+      return "Allowed to create and become a release administrator.";
+    case ALLOWED_ELSA_ADMIN_VIEW:
+      return "Allowed to view as an Elsa-Data administrator.";
+    case ALLOWED_DATASET_UPDATE:
+      return "Allowed to update/refresh dataset index.";
+    default:
+      return `Unknown 'allowed' with code ${allowed}`;
+  }
+}
```
mmalenic: I think it would be a good feature to have a user be able to see their own permissions. I don't think this needs to be just a debug feature? What are you thoughts?

mmalenic (approved):
Looks good! I think the icons with the tool tips are good.

application/frontend/src/pages/users-dashboard/other-users/permission-dialog.tsx
williamputraintan: Yeah, I guess it depends on where the function is being used and whether the re-render component outweigh the  performance of executing the `useCallback`.

There is an [article](https://kentcdodds.com/blog/usememo-and-usecallback#why-is-usecallback-worse) explaining the bad use of `useCallback` (which I thought why I didn't apply the `useCallback`) and a [stack-overflow answer ](https://stackoverflow.com/a/64134704/13137208) which explains it really well.

What I had in mind was `useCallback` needed when we are passing to a custom React `JSX.Element`. I guess the concept should apply if passing it to any component including `<button>`. Happy to wrapped the onSave as a `useCallback` unless we are hitting memory issue (or who knows what) and might start sacrificing re-render.

application/frontend/src/pages/users-dashboard/personal-details-box/personal-details-box.tsx
williamputraintan: Ahh, nice catch! Didn't pay attention enough on the `debug` word there (was part of the existing code). 

Although I am wondering if we should put this as a state/context (and part of API call) so it gets refreshed if permissions was changed. Or maybe just update cookie when needed? I think could be on a seperate PR.

mmalenic: A separate PR makes sense.

I think just updating the cookies for this is the simplest solution, because the cookies could be updated when the permissions change anyway? Although using an api call returning some state also works.

PR #255, opened 2023-03-23
CSRF Token for `<form>` tag

### LINKED ISSUE #233: Lack of CSRF token is preventing Form post for "Download Zip"
We made the "Download Zip" operations to be a special type of form post - in order that the 'save to' download box triggers correctly in the browser. However, this now seems to be failing due to lack of CSRF token.

Need to work out 
- should we just rule these particular routes out of CSRF and what are implications
- should we not do the Form POST technique and deal with download box triggers other way
- can we supply CSRF with Form POST? (best option if possible)

williamputraintan:
CSRF token as an input hidden value field with the name of `_csrf`. This body is now accepted to pass this token as `_csrf` in the body.

### DIFF
--- a/application/frontend/src/components/csrf-token.tsx
+++ b/application/frontend/src/components/csrf-token.tsx
@@ -10,11 +10,9 @@ import { useCookies } from "react-cookie";
  * @returns
  */
 
-function CSRFInputToken() {
+export const CSRFInputToken = () => {
   const [cookies, _setCookie, removeCookie] = useCookies<any>();
   return (
     <input type="hidden" name="_csrf" value={cookies[CSRF_TOKEN_COOKIE_NAME]} />
   );
-}
-
-export default CSRFInputToken;
+};
--- a/application/frontend/src/components/csrf-token.tsx
+++ b/application/frontend/src/components/csrf-token.tsx
@@ -1,0 +1,20 @@
+import { CSRF_TOKEN_COOKIE_NAME } from "@umccr/elsa-constants";
+import React from "react";
+import { useCookies } from "react-cookie";
+
+/**
+ * This is a simple a hidden "_csrf" input that will contain the csrf token.
+ * This will be useful for <form> tag could include csrf token part of the request
+ *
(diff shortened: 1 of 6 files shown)


application/frontend/src/components/csrf-token.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * This is a simple a hidden "_csrf" input that will contain the csrf token.
+ * This will be useful for <form> tag could include csrf token part of the request
+ *
+ * Ref: https://github.com/fastify/csrf-protection#fastifycsrfprotectionrequest-reply-next
+ * @returns
+ */
+
+function CSRFInputToken() {
```
andrewpatto: can I be annoying and get you to do the react components as an "export const"..

I realise they are functionality identical - but it makes the import syntax inconsistent across out components (see below)

application/frontend/src/pages/releases/detail/gcp-storage-iam-share-form.tsx
```diff
@@ -6,6 +6,7 @@ import axios from "axios";
 import { ReleaseTypeLocal } from "../shared-types";
 import { isUndefined } from "lodash";
 import { EagerErrorBoundary, ErrorState } from "../../../components/errors";
+import CSRFInputToken from "../../../components/csrf-token";
```
andrewpatto: if done as an export (no default), then it is consistent between the two imports

application/frontend/src/components/csrf-token.tsx
williamputraintan: Yes, sure!

application/frontend/src/pages/releases/detail/gcp-storage-iam-share-form.tsx
andrewpatto: I mean the advantage of the "default" export is that you can easily rename on import.. but the advantage of the "export const" is that you can have multiples components in a single file

andrewpatto (approved):

PR #256, opened 2023-03-23
Error Unauthorised Response

### LINKED ISSUE #250: Proper Response for Unauthorised Response
Context:
API are now strict to whom are authorised to access specific resource and will return 401/403 HTTP status code. Some component in FE does not handle this properly and kick user out from their login session.

TODO:
Investigate affected components and configure the error boundary appropriately.

williamputraintan:
Seems that 403 with axios interceptor remove user cookie, therefore, it kicks user out from the page. Changing this feature for 401 status code and let 403 is displayed through the ErrorBox component.

In addition:
- hide DAC button from menu bar
- add some ErrorBoundary for DAC page

Fix #250

### DIFF
--- a/application/frontend/src/providers/logged-in-user-provider.tsx
+++ b/application/frontend/src/providers/logged-in-user-provider.tsx
@@ -29,7 +29,7 @@ export const LoggedInUserProvider: React.FC<Props> = (props: Props) => {
     (res) => res,
     (err) => {
       const errCode = err?.response?.status;
-      if (errCode === 403) {
+      if (errCode === 401) {
         removeCookie(USER_SUBJECT_COOKIE_NAME);
       }
       return Promise.reject(err);
--- a/application/frontend/src/layouts/layout-base.tsx
+++ b/application/frontend/src/layouts/layout-base.tsx
@@ -4,6 +4,8 @@ import { useLoggedInUser } from "../providers/logged-in-user-provider";
 import { LayoutBaseHeaderUser } from "./layout-base-header-user";
 import { ErrorBoundary } from "../components/errors";
 import { LayoutBaseFooter } from "./layout-base-footer";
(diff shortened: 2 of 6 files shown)


application/frontend/src/providers/logged-in-user-provider.tsx
```diff
@@ -29,7 +29,7 @@ export const LoggedInUserProvider: React.FC<Props> = (props: Props) => {
     (res) => res,
     (err) => {
       const errCode = err?.response?.status;
-      if (errCode === 403) {
+      if (errCode === 401) {
```
andrewpatto, line 32: I mean the logic surely needs to be more based on the 'type' of auth error.. not just the code?
Specifically - we are removing the cookie when the given cookie JWT is expired? Can we not check for that as a precursor to removing?

andrewpatto (approved):

application/frontend/src/providers/logged-in-user-provider.tsx
andrewpatto: Either through a known error response text (slightly fragile) - or inspecting the JWT (possibly not able to from the front end..).. or we could put the JWT expiry in as a front end readable cookie value (used only for this check)

williamputraintan: Possibly just remove the cookie-removal from the axios interceptor? The cookie itself will expire within a day after its creation anyway?

And also one thing to note that `trpc` is not affected by this I suppose they don't use axios under the hood.

williamputraintan: Possibly have the error boundary to removeCookie when get the desired 401 errors (which could include trpc) and/or have the Fasitfy respond with `clearCookie`

Thinking for next PR ...

PR #259, opened 2023-03-24
Added a new service that calls CloudMap discover

andrewpatto:


andrewpatto:
I have been writing a very AWS specific "Copy Out" service - one that is not appropriate for the general Elsa code base as it only really works with S3.

Anyhow, its kind of like a plugin stack that might be installed/uninstalled.

So we need a dynamic (not cloudformation) mechanism by which functionality like this would be discovered.

application/backend/src/business/services/aws-discovery-service.ts
```diff
@@ -0,0 +1,103 @@
...
+  private copyOutStepsArn: string | null = null;
+
+  /**
+   * Discover the ARN for the Copy Out steps function if it is present in our AWS
+   * setup, or else return null. Aggressively caches the value.
+   */
+  public async locateCopyOutStepsArn(): Promise<string | null> {
+    if (!this.copyOutStepsRunSuccessfully) {
+      const key = "stateMachineArn";
+      const service = "ElsaDataCopyOut";
```
DoxasticFox, line 44: I think the `service` would ideally be configurable, right? This PR LGTM in any case.

DoxasticFox (approved):

application/backend/src/business/services/aws-discovery-service.ts
andrewpatto: Yep - this is very much just a get some infrastructure in place - I'm not sure how the requirements might drive the direction of this so am deferring some better practices until we know more..

PR #262, opened 2023-03-27
A copy out job function

### LINKED ISSUE #261: A Copy Out job that runs as a background task
We need a copy out "service" that can be started as an Elsa background task.

This should take a bucket destination (and or a bespoke cloudformation that the downloader can install) and then in the background transfer all the released files to it.

andrewpatto:
A new way of releasing files involving a massively parallel file copy Steps function.

(that function is implemented elsewhere - but these are the hooks into Elsa for it)


andrewpatto:
Only two participants in release, and only VCFS in AWS

Screenshot 2023-03-27 at 5 26 47 pm

andrewpatto:
Screenshot 2023-03-27 at 5 27 00 pm

Admin says where to send them

andrewpatto:
The job runs

Screenshot 2023-03-27 at 5 28 14 pm

andrewpatto:
Tada..

Screenshot 2023-03-27 at 5 28 27 pm

andrewpatto:
Closes #261

williamputraintan (approved):
It looks nice! ~ LGTM

application/backend/src/business/services/manifests/manifest-bucket-key-helper.ts
```diff
@@ -0,0 +1,74 @@
...
+      return null;
+    };
+
+    for (const art of filesResult.artifacts) {
+      if ("vcfFile" in art) {
+        const c = convert(art["vcfFile"]?.url);
+        if (c) converted.push(c);
+      }
+      if ("bamFile" in art) {
+        const c = convert(art["bamFile"]?.url);
```
DoxasticFox: I was struggling to understand why `_.isString(url)` is called inside of `convert` despite it having the type `url: string`, and I guess it's because the type is actually a lie. `convert` gets passed `art["bamFile"]?.url` which might be `undefined`, if I understand correctly. Ideally, we'd be checking for type errors in CI so that these things don't sneak in.

application/backend/src/business/services/manifests/manifest-bucket-key-helper.ts
```diff
@@ -0,0 +1,74 @@
...
+    const S3_REGEX = new RegExp("^s3://([^/]+)/(.*?([^/]+))$");
+    const GS_REGEX = new RegExp("^gs://([^/]+)/(.*?([^/]+))$");
+    const R2_REGEX = new RegExp("^r2://([^/]+)/(.*?([^/]+))$");
+
+    const convert = (url: string): ManifestBucketKeyObjectType | null => {
+      if (_.isString(url)) {
+        if (url.startsWith(S3_PREFIX)) {
+          const m = url.match(S3_REGEX);
+          if (m) return { service: "s3", bucket: m[1], key: m[2] };
+        }
```
DoxasticFox: The regexes check if the string starts with the prefix, so the `startsWith` check is doing the same thing twice. The URI scheme could also be a capture group. With those changes the body of this loop could be reduced to about 3 lines 🏌️, but if it works, it works

andrewpatto: Yep! Good point - that was a merging of two different code paths and that is redundant now. Will fix.

application/backend/src/business/services/jobs/job-copy-out-service.ts
```diff
@@ -0,0 +1,344 @@
...
+          const describeMapResult = await this.sfnClient.send(
+            new DescribeMapRunCommand({
+              mapRunArn: listMapRunsResults.mapRuns[0].mapRunArn,
+            })
+          );
+
+          // if we have an item count report out the percent done
+          if (describeMapResult.itemCounts) {
+            const countDone =
+              describeMapResult.itemCounts.failed! +
```
DoxasticFox, line 280: Unless I misunderstand, I think it should be easy to avoid disabling the type checker here:

```suggestion
              (describeMapResult.itemCounts?.failed + 0) +
```

etc

application/backend/src/business/services/manifests/manifest-bucket-key-helper.ts
andrewpatto: I was being super defensive - for some reason I can no longer remember.. mainly yes along the lines you said - I think the type that comes via the EdgeDb isn't quite right and doesn't include the "undefined" - where it really can be undefined (so more a bug in the edgedb generator)

application/backend/src/business/services/jobs/job-copy-out-service.ts
```diff
@@ -0,0 +1,344 @@
...
+
+      let manifestCsv = "";
+
+      for (const s of manifest.objects) {
+        // TODO replace with a proper CSV writing component if this becomes an issue (I don't think we are going to hit many
+        //      buckets or keys with quotes)
+        if (s.bucket.includes('"'))
+          throw new Error(
+            `One of the buckets contains a quote character that we cannot handle yet - ${s.bucket}`
+          );
```
DoxasticFox, line 156: While it doesn't hurt to leave check, bucket names can (currently) consist "[only of lowercase letters, numbers, dots (.), and hyphens (-)](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html)."

application/backend/src/business/services/manifests/manifest-bucket-key-helper.ts
```diff
@@ -0,0 +1,74 @@
...
+  const converted: ManifestBucketKeyObjectType[] = [];
+
+  for (const filesResult of masterManifest.specimenList) {
+    // consider moving all these into *UrlService classes..
+
+    const S3_PREFIX = "s3://";
+    const GS_PREFIX = "gs://";
+    const R2_PREFIX = "r2://";
+
+    const S3_REGEX = new RegExp("^s3://([^/]+)/(.*?([^/]+))$");
```
mmalenic, line 24: I'm not sure if this is important, but this regex won't match anything with a trailing `/`. E.g. `s3://bucket/key/`.

DoxasticFox (approved):
Fancy ✨

application/backend/src/business/services/jobs/job-copy-out-service.ts
andrewpatto: The "failed", "total" (etc) properties are required as per the AWS documentation - but for some reason that isn't reflected in the type of itemCount. 

https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-sfn/interfaces/maprunitemcounts.html
vs
https://docs.aws.amazon.com/step-functions/latest/apireference/API_MapRunItemCounts.html

(the ! doesn't disable the type checker - it makes a non-null assertion - the type is then still retained as a number)

application/backend/src/business/services/jobs/job-copy-out-service.ts
mmalenic: This could probably also be relatively easily replaced with something like:
```typescript
import { stringify } from 'csv-stringify/sync';

stringify([s.bucket, s.key]);
```
https://github.com/adaltas/node-csv/tree/master/packages/csv-stringify

Although I don't know if this would stop aws from properly parsing the csv if quotes are escaped.

mmalenic (approved):
Looks good!

application/backend/src/business/services/jobs/job-copy-out-service.ts
andrewpatto: Yep good point - was mainly as a check for Key and was just being symmetric

application/backend/src/business/services/manifests/manifest-bucket-key-helper.ts
andrewpatto: I think we should check that at ingestion level - as we can't currently support the release of a "folder" (which is what the trailing / would imply). That said, that releasing folders is something we might want to do in the future so I will add a suitable comment here..

PR #266, opened 2023-03-28
Reinforce Permission Cookies Authenticity 

### LINKED ISSUE #257: Reinforce Cookie Permissions
Context:
Permission is now listed in Db and could be updated via the API. Cookie might not be the latest permissions when a session has been established.

Todo:
- For BE, should not trust cookie for permissions, but use the permissions listed in Db to grant access for user
- For FE, a mechanism to update ui-cookie when a permission is change from Db.

williamputraintan:
For all internal and `trpc` routes, there will be a permission check with the Db. If these cookie-Db check fail, it will return 401 which led to a redirection to the login page. An alert with a message will display if it happened. This is useful where admin change a user permission where user had logged in with current session.

Manage to attempt modifying UIcookie appropriately for these permissions changed, but `react-cookie` does not detect changes unless using their own `setCookie` function. With this, we also could rely on the information of authUser from `authenticatedRouteOnEntryHelper`.

Also have a chance to play around with [edgedb access policy](https://www.edgedb.com/docs/datamodel/access_policies) which works great in hiding resource from unauthorised user. But, I suppose if we implement this access on the db level, there will be a need for another admin level that could bypass this policy? (e.g. the system will not have a user_uuid to use but need it need to query the Db for a user check OR a powerful user wanted to change somebody else resource) Happy to explore this further, but atm I think this cookie session update should be sufficient for now.

Fix #257 

Screenshot 2023-03-28 at 18 51 25

### DIFF
--- a/application/backend/src/api/errors/authentication-error.ts
+++ b/application/backend/src/api/errors/authentication-error.ts
@@ -2,6 +2,6 @@ import { Base7807Error } from "@umccr/elsa-types";
 
 export class NotAuthorisedCredentials extends Base7807Error {
   constructor(message?: string) {
-    super("Not Authorised With Current Credentials.", 401, `${message ?? ""}`);
+    super("Not Authorised With Current Credentials.", 401, message);
   }
 }
--- a/application/backend/src/api/errors/authentication-error.ts
+++ b/application/backend/src/api/errors/authentication-error.ts
@@ -1,0 +1,7 @@
+import { Base7807Error } from "@umccr/elsa-types";
+
+export class NotAuthorisedCredentials extends Base7807Error {
+  constructor(message?: string) {
+    super("Not Authorised With Current Credentials.", 401, `${message ?? ""}`);
+  }
+}
--- a/application/backend/src/api/helpers/cookie-helpers.ts
+++ b/application/backend/src/api/helpers/cookie-helpers.ts
@@ -1,3 +1,9 @@
+import {
+  ALLOWED_CHANGE_USER_PERMISSION,
+  ALLOWED_CREATE_NEW_RELEASE,
+  ALLOWED_DATASET_UPDATE,
+  ALLOWED_OVERALL_ADMIN_VIEW,
+} from "@umccr/elsa-constants";
 import { FastifyReply, FastifyRequest } from "fastify";
 
 /**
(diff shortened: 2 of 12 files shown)


DoxasticFox (approved):

application/frontend/src/providers/trpc-provider.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const trpcClient = trpc.createClient({
+    links: [
+      httpBatchLink({
+        url: "/api/trpc",
+        async fetch(url, options) {
+          const res = await fetch(url, { ...options });
+
+          const errCode = res.status;
+          if (errCode === 401) {
+            removeCookie(USER_SUBJECT_COOKIE_NAME);
```
andrewpatto, line 21: This still feels very aggressive to me - we have to make sure that under no circumstances anything in the backend actually provides a 401 error to mean anything other than "this person needs to be logged out"..

I mean traditionally I might get a 401 if I attempt to access something I need more permissions for - but it doesn't then aggressively log me out..

application/backend/src/business/services/release-base-service.ts
```diff
@@ -67,7 +67,7 @@ export abstract class ReleaseBaseService {
     user: AuthenticatedUser,
     releaseKey: string
   ) {
-    const isAllowed = user.isAllowedOverallAdministratorView;
+    this.edgeDbClient.withGlobals;
```
andrewpatto: ?

application/backend/src/api/session-cookie-route-hook.ts
```diff
@@ -21,29 +23,26 @@ export function isSuperAdmin(settings: ElsaSettings, user: AuthenticatedUser) {
...
+    if (!dbUser) throw new NotAuthorisedCredentials();
+    request.log.trace(dbUser, `databaseUser: user details`);
 
-      request.log.trace(
-        authedUser,
-        `createSessionCookieRouteHook: user details`
+    // Check for permissions different
+    const dbPermission = createUserAllowedCookie(dbUser);
+    const sessionPermission = createUserAllowedCookie(authedUser);
+    if (dbPermission != sessionPermission) {
```
andrewpatto, line 39: whilst you will probably be ok - I'm not sure there is any implied Set ordering.. so theoretically the string created from two calls to Array.from(Set<string>) could end up in different orders and hence fail this check..

application/backend/src/api/helpers/cookie-helpers.ts
```diff
@@ -41,3 +47,46 @@ export function cookieForBackend(
...
+
+  if (isAllowedOverallAdministratorView) {
+    allowed.add(ALLOWED_OVERALL_ADMIN_VIEW);
+  }
+
+  if (isAllowedChangeUserPermission) {
+    allowed.add(ALLOWED_CHANGE_USER_PERMISSION);
+  }
+
+  return Array.from(allowed.values()).join(",");
```
andrewpatto, line 91: possibly a sort here if you are relying on consistent ordering..

application/backend/src/api/errors/authentication-error.ts
```diff
@@ -0,0 +1,7 @@
+import { Base7807Error } from "@umccr/elsa-types";
+
+export class NotAuthorisedCredentials extends Base7807Error {
+  constructor(message?: string) {
+    super("Not Authorised With Current Credentials.", 401, `${message ?? ""}`);
```
andrewpatto: I would check whether the underlying Base7807 requires the field - and if it does - not sure an empty string helps.. and if not, then not sure I'd pass anything if message is empty

andrewpatto (approved):

andrewpatto:
Looks good - happy to give this a go - I feel I'll understand more once I see it running in normal operations. I still have some concerns about the exact semantics of some of the operations but that is around the edges..

application/backend/src/api/session-cookie-route-hook.ts
andrewpatto: I tell a lie "The iteration over a set visits elements in insertion order." - so should be fine

application/backend/src/business/services/release-base-service.ts
williamputraintan: Ahh, I missed this, was exploring on global on edgedb. Thanks

application/frontend/src/providers/trpc-provider.tsx
williamputraintan: Hmmm, I am thinking such a scenario would result in a 403? The creds is still valid but the app is refusing bcause of smth else?

I do agree parsing errCode might be a bit tricky, I guess could reserve other errorCode for this? Just thinking to represent some kind of error on every request (including `curl`) when user permissions changed (user need to relogin).

andrewpatto: Happy with that - but it is easy for someone to throw a 401 thinking it is much the same as a 403.. (I'm just saying we are going to have to be very rigorous that we understand the ramifications of using a 401 in our server code)

PR #268, opened 2023-03-29
Generate TSV manifest from master manifest

DoxasticFox:
Key points:

* The "MANIFEST" tab in the "Access Data" pane produces a TSV manifest now.
* Presigned URLs are generated at manifest creation time, and added to the TSV if requested.
* The "PRESIGNED URLS" tab has been removed from the "Access Data" pane.
* Manifests can only be generated if `userRole === "Manager" || userRole === "Member"`.


application/backend/src/api/routes/internal/release-routes.ts
```diff
@@ -35,6 +36,7 @@ export const releaseRoutes = async (fastify: FastifyInstance) => {
     ReleaseParticipationService
   );
   const releaseSelectionService = container.resolve(ReleaseSelectionService);
+  const manifestService = container.resolve(ManifestService);
```
DoxasticFox: The merge conflict is because of this line. I have my branch based on an out-of-date commit because the latest one broke `origin/dev`. I might wait for @andrewpatto to fix it before merging this PR

andrewpatto:
Looks good..

I wonder if we might have a discussion down the line (with AG) as to the usefulness of them being able to get manifests that *don't* contain presigned URLS, but do contain other data. That was the original intention of the split between "manifest" tab and "presigned URLS" tab - but I think I agree with your approach here of unifying. I just wonder whether in the case they don't have permission to the files themselves (so are release admins or viewers) - that we just gray out a set of object fields that they aren't allowed.

I think the use case I am thinking about here is they get contacted about "where is file XYZ" - and they could go into a release and at least get the list of objects in the manifest - if not the data itself.

(for a future PR - this is fine)

andrewpatto (approved):

williamputraintan (approved):

application/frontend/src/pages/releases/detail/manifest-form.tsx
```diff
@@ -0,0 +1,69 @@
...
+      id={`chx-${field}`}
+      value={field}
+    />
+    <label className="uppercase" htmlFor={`chx-${field}`}>
+      {field}
+    </label>
+  </div>
+);
+
+export const ManifestForm: React.FC<Props> = ({ releaseKey, releaseData }) => {
```
mmalenic, line 30: Should this form be hidden if the user is an `Administrator`? Is there a reason why `Administrators` can't access the manifest?

mmalenic (approved):

application/frontend/src/pages/releases/detail/manifest-form.tsx
DoxasticFox: I have no idea lol, so I'm going to defer to @williamputraintan and @andrewpatto on that. Maybe only certain checkboxes should be hidden?

Irrespective of whether there should be changes, I'm going to merge this PR ASAP because there's some schema changes which block the other PRs that are open right now

PR #200, opened 2023-02-10
`csrf-token` mechanism

williamputraintan:
- Implement CSRF token for internal routes, which requires `csrf-token` header on requests
- Token stored in the client cookie and use the `axios.interceptor` to include as header

PR #269, opened 2023-03-29
Implement csrf-token in TRPC

williamputraintan:
Patch #200

### DIFF
--- a/application/backend/src/api/api-trpc-routes.ts
+++ b/application/backend/src/api/api-trpc-routes.ts
@@ -1,0 +1,17 @@
+import { FastifyInstance } from "fastify";
+import { fastifyTRPCPlugin } from "@trpc/server/adapters/fastify";
+import { appRouter } from "../app-router";
+import { createContext } from "./routes/trpc-context";
+
+/**
+ * Defined for trpc routes
+ *
+ * @param fastify
+ */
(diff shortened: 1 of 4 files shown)


application/backend/src/api/api-trpc-routes.ts
```diff
@@ -0,0 +1,17 @@
...
+import { appRouter } from "../app-router";
+import { createContext } from "./routes/trpc-context";
+
+/**
+ * Defined for trpc routes
+ *
+ * @param fastify
+ */
+export const trpcRoutes = async (fastify: FastifyInstance) => {
+  fastify.addHook("preHandler", fastify.csrfProtection);
```
williamputraintan, line 12: Thinking to make use of `fastify.addHook` for csrf-token checking, and so breaking this to routes before registering the `fastifyTRPCPlugin`.

andrewpatto:
I guess I haven't thought too much about it - is it even possible for a CSRF exploit over TRPC?

andrewpatto:
I mean I guess a lot of the "payload" for TRCP goes into the URL itself and therefore could be CSRF exploitable..

williamputraintan:
I suppose yes, I guess it is just a more longer url for the exploitation ? aaand it is now equal regardless the internal api uses trpc or not both have csrf-token protection.

andrewpatto (approved):

PR #270, opened 2023-03-29
A child dependency injection container per request

andrewpatto:
Setup for each and every TRPC and REST request.
This is the way I should have set up DI from the start - but instead we were using the global container.


andrewpatto:
The use of "dc" rather than "container" as a variable name is to prevent the accidental import of "container" (the global) which jetbrains will for instance helpfully import as you paste the field container into code..

application/backend/src/bootstrap-dependency-injection.ts
```diff
@@ -33,37 +49,43 @@ export function bootstrapDependencyInjection() {
...
   });
-  container.register<IPresignedUrlProvider>("IPresignedUrlProvider", {
+  dc.register<IPresignedUrlProvider>("IPresignedUrlProvider", {
     useClass: CloudflarePresignedUrlsService,
   });
+
+  // we register our singletons this way as this is the only way to prevent them being registered
+  // in the global container namespace (we DON'T use the @singleton decorator)
+  dc.registerSingleton(AwsDiscoveryService);
+
```
andrewpatto, line 89: of note - we cannot use @singleton as a decorate as that *always* registers into the global container - and the global container will not have an ability to resolve services like "Database" (as we introduce these one child container down).
So we will need to be explicit about any particular service lifespans here

application/backend/src/api/routes/internal/release-routes.ts
```diff
@@ -17,24 +17,30 @@ import {
...
 
-export const releaseRoutes = async (fastify: FastifyInstance) => {
-  const presignedUrlsService = container.resolve(PresignedUrlsService);
-  const awsAccessPointService = container.resolve(AwsAccessPointService);
-  const gcpStorageSharingService = container.resolve(GcpStorageSharingService);
-  const releaseService = container.resolve(ReleaseService);
-  const releaseParticipantService = container.resolve(
+export const releaseRoutes = async (
+  fastify: FastifyInstance,
+  _opts: { container: DependencyContainer }
```
andrewpatto, line 30: The routes now are _more_ correct in that they get passed in a container.. *eventually* though we should get these to resolve services in the request code path - not in the route code path (but can defer that for a bit)

andrewpatto:
It looks like a lot of changes - but is mainly
- removing @singletons
- renaming container to dc
- removing any reference to the global container (need to pass it in in some places like test generators)

andrewpatto:
Had to rewire some of the tests so that they got DI containers more correctly

application/backend/src/business/services/aws-cloudtrail-lake-service.ts
```diff
@@ -38,7 +37,7 @@ type CloudTrailLakeResponseType = {
 };
 
 @injectable()
-@singleton()
+@scoped(Lifecycle.ContainerScoped)
```
mmalenic: Should these all be `@scoped(Lifecycle.ContainerScoped)` or is it just this aws service?

application/backend/src/business/services/aws-discovery-service.ts
```diff
@@ -16,7 +16,6 @@ import {
  * once other deployment environments exist.
  */
 @injectable()
-@singleton()
```
mmalenic, line 19: See above [comment](https://github.com/umccr/elsa-data/pull/270#discussion_r1152624532).

mmalenic (approved):
Looks good! I think it makes sense not to use the global container. This might also allow reusing some tokens, e.g. if some services need an `AWS`, or `GCP` version of a dependency that's implemented behind a common interface.

application/backend/src/business/services/aws-cloudtrail-lake-service.ts
andrewpatto: Yes good catch - I think this was an experiment I left in.. I will remove

application/backend/src/app.ts
```diff
@@ -14,43 +14,58 @@ import {
...
+
     this.server.addHook("onRequest", async (req, reply) => {
       // we make a shallow copy of the settings on each fastify request to (somewhat)
       // protect against routes doing mutations
       (req as any).settings = { ...settings };
+      // give each request its own DI container
+      (req as any).container = dc.createChildContainer();
+    });
+
+    // similarly for TRPC, start each request context with a copy of the Elsa settings and a custom child DI container
```
williamputraintan, line 65: I think we are missing the `req` and `res` right? `ctx` downstream seems to be missing these values.

```suggsted
    this.trpcCreateContext = async (opts: CreateFastifyContextOptions) => ({
      settings: { ...settings },
      container: dc.createChildContainer(),
      req: opts.req,
      res: opts.res,
    });
```

If so I could add this next PR

PR #273, opened 2023-04-02
bug(backend): fix req and res not being included in the tRPC router context

mmalenic:
Fixes the missing response and request context in `trpcCreateContext`.

This is just so that we don't all implement that same fix in different PRs. Based on https://github.com/umccr/elsa-data/pull/270#discussion_r1152787524.

### DIFF
--- a/application/backend/src/app.ts
+++ b/application/backend/src/app.ts
@@ -20,10 +20,10 @@ import { Logger } from "pino";
 import { apiExternalRoutes } from "./api/api-external-routes";
 import { apiUnauthenticatedRoutes } from "./api/api-unauthenticated-routes";
 import { getMandatoryEnv, IndexHtmlTemplateData } from "./app-env";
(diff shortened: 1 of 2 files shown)


DoxasticFox (approved):
🙌

williamputraintan (approved):

application/backend/src/app.ts
```diff
@@ -63,9 +65,11 @@ export class App {
     });
 
     // similarly for TRPC, start each request context with a copy of the Elsa settings and a custom child DI container
-    this.trpcCreateContext = async () => ({
+    this.trpcCreateContext = async (opts: CreateFastifyContextOptions) => ({
```
andrewpatto, line 68: ah, this makes more sense.. I was kind of thinking that they were automatically injected into the context *after* the initial createContext. Good fix.

andrewpatto (approved):

Issue #144, opened 2022-12-06
Create an external API endpoint that returns a 'release' manifest


andrewpatto:
We have endpoints for generating TSV manifests etc - but all are set up for our internal frontend->backend API (authenticated with session cookies etc).

We also want an API surface that is useable by external parties - and where we will give stronger guarantees of backward compatibility.

First cab off the rank should be an external API that allows the manifest data for a release to be retrieved in JSON.

(this will be useful for say a htsget endpoint)

mmalenic:
By manifest data for a release, do you mean data that is fetched by this api call? https://github.com/umccr/elsa-data/blob/fc947801b18b1b465cfd4ad92c060f534a5ddbcd/application/backend/src/api/routes/release-routes.ts#L418-L451

What would the authentication look like for this? Should it use the same information as the openid login system?

andrewpatto:
Yes - similar in spirit to this - but can contain whatever is needed by htsget in whatever format it needs.

The manifest from that endpoint is TSV just because that's how bioinformaticians like things.

I imagine something that almost mirrors the underlying Edgedb dataset case model

```
[
  { id: "SINGELTONMARY", patients: [
    { id: "MARY": specimens: [
       { id: "HG0001", vcfUrl: "s3:///", bamUrl: "s3:///...", geneRegionsAllowed: [ "HGNC:123", "HGNC:3535" ] }
     ] )
 ]
]
```

For authentication - I'm happy for literally anything at the moment. Seriously, B2b pre shared keys would be fine..

Mutual TLS?

The assumption from Elsa is that htsget is a good actor and that htsget is doing whatever auth it needs.. I mean all Elsa is doing is providing the list of files and id - htsget already has permission to access the data files so its not getting that permission from Elsa

mmalenic:
Yeah that's fair enough. But Elsa would still want to make sure that it is sending the information to htsget and not something else right? Although I guess that doesn't need to be anything super advanced.

andrewpatto:
Yeah that's why something like mutual TLS would be the strongest - as both ends are proving to each other who they are.

But shared secret is fine for the moment

PR #274, opened 2023-04-03
feat: htsget manifest route

mmalenic:
Related to https://github.com/umccr/elsa-data/issues/144

This PR implements (changes the existing one) a public, unauthenticated route which returns a manifest for use by htsget. The route simply uploads the manifest to a bucket, and returns the bucket and object key for where it is located. It is expected that htsget will have permissions to get this manifest file, and that it can then use it to map queries for particular releases.

### Changes
* Add unauthenticated htsget manifest route, which uploads a manifest file to an S3 bucket. This file can be cached to avoid re-uploading unnecessarily.
* Add an interface that defines put and head requests for cloud storage. Storage can be implemented for other cloud providers like GCP, or Cloudflare using this interface.
* Add an `isAllowedHtsget` to the releases schema.
* The route can be enabled/disabled in the releases page.
* Configuration for the route consists of a `manifestTTL` which indicates the how long the manifest is cached for, and the htsget url.
* Add some tests for this functionality.


application/backend/src/business/services/aws/aws-enabled-service.ts
```diff
@@ -0,0 +1,49 @@
+import { GetCallerIdentityCommand, STSClient } from "@aws-sdk/client-sts";
+import { inject, injectable } from "tsyringe";
+import { Logger } from "pino";
+
+/**
+ * The AWS enabled service detects that are enough permissions to use
+ * AWS.
+ */
+@injectable()
+export class AwsEnabledService {
```
mmalenic, line 10: I think this is a better way to represent a `base-service`. This way the base service is composable rather than using inheritance, which allows the `enabled` check to run only once in a singleton. I've also moved the `enabled` check into an awaitable function, which means that the first time the check is called, it won't throw an error until it gets a result.

application/backend/src/business/services/gcp-storage-sharing-service.ts
```diff
@@ -96,6 +95,8 @@ export class GcpStorageSharingService extends GcpBaseService {
         throw Error(`Unexpected operation ${operation}`);
       }
 
+      await this.gcpEnabledService.enabledGuard();
+
```
mmalenic: Another way to do these `enabledGuard` checks would be to write a decorator that calls the function before each method in a class, as described in this [answer](https://stackoverflow.com/a/49120018). That way it cannot be forgotten, however it does have the disadvantage of applying to all methods in a class, and some methods might not need this check.

application/backend/src/api/bearer-route-hook.ts
```diff
@@ -4,11 +4,11 @@ import { UsersService } from "../business/services/users-service";
 export function createBearerRouteHook(usersService: UsersService) {
   return async (request: FastifyRequest, reply: FastifyReply) => {
     try {
+      // Should these be completely unauthorised?
```
mmalenic, line 7: I'm not sure if this should be _completely_ unauthorised, but I think it is okay because it returns buckets and keys that need external permissions anyway.

application/backend/config/base.json5
```diff
@@ -6,4 +6,7 @@
   oidc: {
     issuerUrl: "https://test.cilogon.org",
   },
+  htsget: {
+    url: "https://htsget.elsa.dev.umccr.org/",
```
andrewpatto: is there a reason why the URL needs to be trailing slash?

application/backend/src/api/api-auth-routes.ts
```diff
@@ -240,6 +239,13 @@ export const apiAuthRoutes = async (
           await reply.generateCsrf()
         );
 
+        cookieForUI(
+          request,
+          reply,
+          USER_HTSGET_URL,
+          settings.htsget?.url.toString()
+        );
+
```
andrewpatto: I don't fundamentally hate it - but is there a reason a data field like this becomes "cookie" worthy? (as opposed to just being a field passed down from GET of Release?

application/backend/src/api/bearer-route-hook.ts
andrewpatto: Agreed - look something make come out of all of the GA4GH auth discussions that help out here - for the moment this is a perfectly good local/extensible/quick solution

application/backend/config/base.json5
mmalenic: No reason, just left it in accidentally.

application/backend/src/api/api-auth-routes.ts
mmalenic: The cookie was just a simple way to pass the data to the front end, but I agree, I think this is better as a field when the release info is fetched.

application/backend/src/api/helpers/cookie-helpers.ts
```diff
@@ -18,8 +18,12 @@ export function cookieForUI(
   request: FastifyRequest,
   reply: FastifyReply,
   k: string,
-  v: string
+  v?: string
 ) {
+  if (v === undefined) {
+    return;
+  }
+
```
andrewpatto: Are we sure this matches the semantics of how the front cookie get/check works?

application/backend/src/business/services/aws/aws-enabled-service.ts
andrewpatto: yes - good call

application/common/elsa-types/schemas-releases.ts
```diff
@@ -90,6 +83,9 @@ export const ReleaseDetailSchema = Type.Object({
   isAllowedGSData: Type.Boolean(),
   isAllowedR2Data: Type.Boolean(),
 
+  isAllowedHtsget: Type.Boolean(),
+  isHtsgetEnabled: Type.Boolean(),
+
```
andrewpatto, line 94: can we explain the differences between these two in (short) comments here.. (its just going to be non-obvious to someone browsing what the distinction is)

application/backend/src/config/config-constants.ts
```diff
@@ -72,6 +72,30 @@ export const configDefinition = {
       env: `${env_prefix}SESSION_SALT`,
     },
   },
+  htsget: {
+    enabled: {
```
andrewpatto: We can't infer this just from the presence of the `htsget` section of the config? What are the scenarios were someone would want to fill in the htsget URL but not enable it?

application/backend/src/business/services/release-base-service.ts
```diff
@@ -83,6 +85,16 @@ export abstract class ReleaseBaseService {
...
 
+  /**
+   * Is a feature enabled in the config. Currently, this can just check for htsget.
+   * It might be good to check this for other features that require config properties to
+   * be set.
+   * @param property
+   */
+  public isFeatureEnabled(property: "isAllowedHtsget"): boolean {
+    return this.settings.htsget !== undefined;
+  }
```
andrewpatto, line 99: I like this pattern so keep up with this - not sure it needs an actual boolean in the config (which by the way this one seems not to check - it just checks the existence of the config section?)

application/backend/src/business/services/gcp-storage-sharing-service.ts
andrewpatto: typescript 5 has literally just two weeks ago added a new decorator syntax that will be more compatible with javascript decorators going forward - happy to try this out soon - but was waiting for all the tooling around typescript 5 to settle first

andrewpatto: https://devm.io/typescript/typescript-5-decorators

application/backend/config/base.json5
andrewpatto: Given the URL has to participate in URL fiddling - I think I prefer it doesn't come with a trailing slash - and that we actually enforce that at the config level. That way we have simpler construction rules throughout when we need to make URL+path (I realise there are libraries that probably safely do URL construction with or without slashes so it probably doesn't matter - but still - mind as well be dogmatic at the config level)

mmalenic:
Update: I think I'll also add a TTL or max age to the response to indicate how long it is valid for. Perhaps this can be in a Cache-Control header, rather than having a `cached` field.

andrewpatto (approved):

andrewpatto:
> Update: I think I'll also add a TTL or max age to the response to indicate how long it is valid for. Perhaps this can be in a Cache-Control header, rather than having a `cached` field.

Yes - would prefer that. I even made a "manifest" etag on construction of the manifest - so possibly some cache shenanigans could be done with that (the htsget endpoint can send in the etag that it currently has cached and we can tell it "keep going" if that is still the manifest tag) (there are standard cache-control headers for this).

But equally just some time based ones would be good.

Even a combination maybe

application/backend/src/business/services/release-base-service.ts
mmalenic: I've simplified this so that there is no `htsget.enabled` flag in the config, instead it is inferred from the presence of `htsget.url`.

application/backend/src/business/services/audit-log-service.ts
```diff
@@ -799,46 +822,80 @@ export class AuditLogService {
...
       const errorString =
         error instanceof Error ? error.message : String(error);
 
-      await this.completeReleaseAuditEvent(
+      await completeAuditFn.call(
+        this,
         this.edgeDbClient,
         auditEventId,
-        OUTCOME_SERIOUS_FAILURE,
+        OUTCOME_SUCCESS,
```
williamputraintan, line 898: Is this intended to become success while it is actually a catch from an error? If you could recall this 😅 @mmalenic 

I was looking at the auditPattern and [GitLens](https://marketplace.visualstudio.com/items?itemName=eamodio.gitlens) shows me this PR for this line of change.

mmalenic: I think that should be `OUTCOME_SERIOUS_FAILURE`, because it's processing the failed error state! Good catch, I must have accidentally changed it when refactoring this code.

mmalenic: Although I guess it depends on what you're trying to do. Maybe on an error, in some contexts it should only be a `OUTCOME_MINOR_FAILURE`? Perhaps this function should have a few varieties, depending on which outcome an error state can cause?

williamputraintan: Yeah agree need some error parsing there whether it belongs to SERIOUS or MINOR error.

williamputraintan: I'll just put this as an issue for now https://github.com/umccr/elsa-data/issues/449

PR #278, opened 2023-04-05
Just small tidies on the log messages

andrewpatto:


andrewpatto:
Our logs from the job handler were a bit busy for very little value..
And just some other places we were printing rather than logging

DoxasticFox (approved):

williamputraintan (approved):

mmalenic (approved):

PR #279, opened 2023-04-06
E2E test script for create manual release -> edit cases

### LINKED ISSUE #232: Do a E2E test script for create manual release -> edit cases
Am finding in-between things breaking (i.e. the handoff from create manual release dialog to view the release now redirects to the wrong URL).. which is all good - not hard to fix - but we need some more UI testing that will detect these. So lets start with a simple "Create / View / Edit" flow

DoxasticFox:
Closes https://github.com/umccr/elsa-data/issues/232.

`npx playwright codegen` and `npx playwright test --trace on` are pretty good

### DIFF
--- a/e2e/playwright.config.ts
+++ b/e2e/playwright.config.ts
@@ -60,7 +60,8 @@ const config: PlaywrightTestConfig = {
   /* set up so our test server runs directly from dev */
   webServer: {
     cwd: "../application/backend",
-    command: "node --loader ts-node/esm src/entrypoint.ts",
+    command:
+      "node --loader ts-node/esm src/entrypoint.ts web-server-with-scenario 1",
     port: 3000,
     // there should be nothing running on our port in CI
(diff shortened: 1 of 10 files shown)


e2e/playwright.config.ts
```diff
@@ -60,7 +60,9 @@ const config: PlaywrightTestConfig = {
   /* set up so our test server runs directly from dev */
   webServer: {
     cwd: "../application/backend",
-    command: "node --loader ts-node/esm src/entrypoint.ts",
+    command:
+      "node --loader ts-node/esm src/entrypoint.ts web-server-with-scenario 1",
```
andrewpatto, line 64: well that explains why they randomly broke in a non-deterministic way!

andrewpatto:
excellent start - @williamputraintan  and @mmalenic - you are the next cabs off the rank for doing a E2E test..

andrewpatto (approved):

PR #285, opened 2023-04-13
Attempting to slim the docker using esbuild

### LINKED ISSUE #243: Slim the Elsa Data docker image
We use builders to make sure that we don't need the whole EdgeDb database installed etc..

HOWEVER, I paid no attention to actually what is needed in the base deployable Elsa image. I think it is currently like 1Gb because it is just a standard AWS image. There are lots of techniques (alpine nodejs?) for slim deployable node images.

I'd concentrate on the base first..

THEN - I guess we can also avoid installing all the source code and node_modules (we literally do COPY . .) 

BUT I wouldn't prioritise this.. that can be something for way down the track.. 

My guess is we can get it down to 200MB just by selecting a better base image

andrewpatto:
In draft just so I can get some published Docker images for testing

This is all still entirely experimental


andrewpatto:
Ok I've basically got it working - but there could still easily be unforseen breakages - so I'm not declaring this fully working yet. Lets merge and keep monitoring

application/Dockerfile
```diff
@@ -60,125 +60,82 @@ RUN edgedb-server \
...
+# Deployed app
+# Optimised for size
+#
+FROM node:18-alpine
+
+# we want to embed into the image environment immutable tags passed in by the build environment
 # this is the 'semantic' version as set by developers i.e. 1.0.3
 ARG ELSA_DATA_VERSION
-RUN ["/bin/bash", "-c", ": ${ELSA_DATA_VERSION:?Build argument needs to be set and not null.}"]
+RUN ["/bin/ash", "-c", "[ -n \"${ELSA_DATA_VERSION}\" ]"]
```
andrewpatto: alpine shell is only one available

application/backend/bundle-build.js
```diff
@@ -0,0 +1,14 @@
...
+build({
+  entryPoints: ["src/entrypoint.ts", "jobs/entrypoint-job-handler.ts"],
+  external: ["sodium-native", "node-gyp-build"],
+  bundle: true,
+  outdir: "server/dist",
+  platform: "node",
+  minify: true,
+  plugins: [esbuildPluginPino({ transports: ["pino-pretty"] })],
+  target: "node18",
+}).catch(() => process.exit(1));
```
andrewpatto, line 16: we need to use a more complex esbuild technique so we can use the special plugin for 'pino'

application/backend/src/api/routes/internal/audit-event-routes.ts
```diff
@@ -19,7 +19,7 @@ import {
 import { ElsaSettings } from "../../../config/elsa-settings";
 import _ from "lodash";
 import { DependencyContainer } from "tsyringe";
-import AuditEvent = audit.AuditEvent;
+import AuditEvent = interfaces.audit.AuditEvent;
```
andrewpatto, line 22: some wierd esbuild quirk workaround

application/backend/src/entrypoint-command-start-web-server.ts
```diff
@@ -95,11 +97,22 @@ export async function startWebServer(
  * @param config
  */
 export async function startJobQueue(config: any) {
+  let jobFileName = "entrypoint-job-handler.ts";
+  let root = path.resolve("jobs");
```
andrewpatto, line 101: a way to distinguish between the Bree setup for local dev v deployed

andrewpatto:
Screenshot 2023-04-13 at 4 14 26 pm

andrewpatto:
Final size is 196MB, down from 900+

mmalenic (approved):
Nice!

andrewpatto:
Actually its blown out to 283MB - I forgot that we need to run the `edgedb` binary to do migrates etc - so I can't use alpine. Using debian buster slim. 283MB is still fine..

PR #286, opened 2023-04-14
Lowered log for health check

andrewpatto:
Also split the health check routes and added some small details

### DIFF
--- a/application/backend/src/api/routes/unauthenticated/health.ts
+++ b/application/backend/src/api/routes/unauthenticated/health.ts
@@ -2,6 +2,7 @@ import { FastifyInstance } from "fastify";
 import { DependencyContainer } from "tsyringe";
 import { getServices } from "../../../di-helpers";
 import { version, platform } from "node:process";
+import { getMandatoryEnv } from "../../../app-env";
 
 /**
  * Routes that exist mainly for confirming functionality (health checks).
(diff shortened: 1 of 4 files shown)


andrewpatto:
AWS is pretty aggressive on the health checks - but it means out cloudwatch is completely filled with health checks logs

andrewpatto:
@DoxasticFox not sure if you were using the health check endpoint in the Google deployment - but if so it has changed path!

application/backend/src/api/routes/unauthenticated/health.ts
```diff
@@ -0,0 +1,64 @@
...
+  fastify.get("/health/check-detailed", {}, async function (request, reply) {
+    const { edgeDbClient, logger } = getServices(opts.container);
+    // do a log very early (just in case we have a problem with responding - this at least lets us know the input
+    // got to the backend)
+    logger.info("health-check-detailed: called");
+
+    // confirm our database is still accessible
+    let dbResult = "";
+    try {
+      const fortyTwoResult = await edgeDbClient.query("select 42;");
```
mmalenic, line 46: I would've been boring and done `select true;` to indicate `isDatabaseUp`, but I like this.

mmalenic (approved):

PR #288, opened 2023-04-14
TRPC for Dataset

### LINKED ISSUE #281: Convert the datasets API/react router surface to use their URI rather than db id
We currently show the Dataset database id in a couple of spaces - when in fact the true 'id' for each dataset is the externally defined Uri. We should use this Uri everywhere including the React routing and API.

This is a mirror of what we already did for Releases

williamputraintan:
- Move dataset-summary query to `.edgeql`
- Move internal `/dataset` to `trpc-dataset`
- use `datasetUri` as main ID instead of `datasetDbId`


application/backend/package.json
```diff
@@ -64,8 +64,8 @@
     "@google-cloud/secret-manager": "^4.2.1",
     "@google-cloud/storage": "^6.9.2",
     "@sinclair/typebox": "0.25.10",
-    "@trpc/client": "10.18.0",
-    "@trpc/server": "10.18.0",
```
williamputraintan, line 68: There are some types error that can't be imported in frontend using this version but works when downgrading this.

Screenshot 2023-04-14 at 20 18 55

andrewpatto: is there anything in the change history for the trpc libraries that would indicate what has changed?

williamputraintan: Managed to update to `10.20.0` on both trpc FE and BE seems to be working now.

Couldn't see obvious changes on the change log but it seems that both `trpc-react` and `trpc-client/server` should have the same version.

e2e/tests/dac-manual-release.spec.ts
```diff
@@ -18,9 +18,22 @@ test("A release can be created manually which has selectable cases", async ({
...
+
+  await page
+    .locator("#datasetUriSelector > div")
+    .getByText("urn:fdc:umccr.org:2022:dataset/10f")
+    .click();
+
+  await page
+    .locator("#datasetUriSelector > div")
+    .getByText("urn:fdc:umccr.org:2022:dataset/10g")
+    .click();
```
DoxasticFox, line 35: Nice

application/frontend/src/pages/releases/detail/cases-box/consent-summary.tsx
```diff
@@ -84,29 +85,22 @@ function ConsentSummary({ consentId, releaseKey, nodeId }: Props) {
...
+    consentQuery = trpc.datasetRouter.getDatasetConsent.useQuery({ consentId });
   } else {
-    u = `/api/releases/${releaseKey}/consent/${nodeId}`;
+    consentQuery = trpc.releaseRouter.getReleaseConsent.useQuery({
+      releaseKey: releaseKey ?? "",
+      nodeId: nodeId ?? "",
+    });
   }
 
+  const duosCode = (consentQuery.data ?? []) as DuoLimitationCodedType[];
```
DoxasticFox: This is feedback only on style, so feel free to ignore: Type annotations are preferable to type assertions:

```suggestion
  const duosCode: DuoLimitationCodedType[] = consentQuery.data ?? [];
```

Annotations give TypeScript the opportunity to check the types' correctness but assertions don't. It doesn't make any difference in this case, because TypeScript doesn't know what `consentQuery.data` is (it thinks it's `any`), but if the code were ever refactored such that `consentQuery.data`'s type was something else, it's be nice to give TypeScript the opportunity to call-out potential problems

application/frontend/src/components/dataset/dataset-table.tsx
```diff
@@ -0,0 +1,154 @@
...
+                <td className="whitespace-normal break-words">
+                  <div className="font-bold"> {row.description}</div>
+                  <div className="flex flex-row space-x-2 text-sm">
+                    <p className="font-mono opacity-50">{row.uri}</p>
+                  </div>
+                </td>
+
+                <td className={classNames(baseColumnClasses, "text-left")}>
+                  {row.updatedDateTime
+                    ? formatLocalDateTime(row.updatedDateTime as string)
```
DoxasticFox: Type-safe alternative (which I'm not sure is even necessary, because by IDE says `row.updatedDateTime` is `any`):

```suggestion
                    ? formatLocalDateTime(String(row.updatedDateTime))
```

DoxasticFox (approved):
LGTM

application/backend/dbschema/queries/dataset/getDatasetSummary.edgeql
```diff
@@ -0,0 +1,192 @@
...
+    userPermission.isAllowedOverallAdministratorView
+  ),
+
+  dataset := (
+    select dataset::Dataset
+    filter
+      (
+        .uri = <str>$datasetUri
+          or
+        (not <bool>$isSingleUriQuery)
```
mmalenic: You might be able to represent this by having an `<optional str>$datasetUri`, and checking if the the parameter is set by querying it with the `exists` operator, e.g. `select exists <optional str>$datasetUri;`. An empty optional should evaluate to false with `exists`. That way you don't need two variables to represent this, which would simply calling the query.

mmalenic (approved):
Looks good!

application/backend/dbschema/queries/dataset/getDatasetSummary.edgeql
williamputraintan: Ahh, yes, thank you! Was looking for the `optional` param there but perhaps I miss this 🤦‍♂️.

andrewpatto (approved):

Issue #271, opened 2023-03-30
Rescue the service tests including splitting out to cloud v non-cloud


andrewpatto:
We currently don't run the service tests in github actions - mainly because some of the service tests use AWS infrastructure and we haven't set up a way for that to run in github (or without the actual cloud)

But that allows some of the service tests to drift (I know I broke some in a recent PR for instance - but only discovered now).

I think we should both fix the service tests - but also split out those that can't run in github (and I'm prepared to let them be a local only test for the moment)

andrewpatto:
We then should run the service tests as part of the github testing

andrewpatto:
A lot of the "release" tests have broken I think because of the new permissions work - and because of the new manifest work. Fixing them is mainly just about correcting the "counts" returned etc

andrewpatto:
Actually maybe Christian as the person who has least access to the UMCCR AWS - is possibly in a good position to work out how best to make this work without real cloud permissions..

andrewpatto:
[Localstack?
](https://localstack.cloud/)

PR #299, opened 2023-04-21
refactor(backend): add manual inject tokens to all injectable classes

mmalenic:
Esbuild is not as compatible with tysringe as ts-node; specifically it does not support `emitDecoratorMetadata`. Tsyringe can still be used, but constructor dependencies must use manually injected tokens (e.g. `@inject(Dependency) dependency`), otherwise that dependency is undefined. 
* Change all class dependencies to manually use @inject.
* Add `private` and `readonly` to class properties where possible.
* Rename plural services to singular.
* Use esbuild-jest instead of ts-jest so that esbuild errors are caught in tests.

### DIFF
--- a/application/backend/tests/service-tests/client-mocks.ts
+++ b/application/backend/tests/service-tests/client-mocks.ts
@@ -1,0 +1,19 @@
+import { mockClient } from "aws-sdk-client-mock";
+import { GetCallerIdentityCommand, STSClient } from "@aws-sdk/client-sts";
+import { injectable } from "tsyringe";
+
+/**
+ * Mock the AwsEnabledService.
+ */
+@injectable()
+export class AwsEnabledServiceMock {
+  private readonly stsClientMock = mockClient(STSClient);
+
+  public enable() {
+    this.stsClientMock.on(GetCallerIdentityCommand).resolves({});
+  }
+
+  public reset() {
+    this.stsClientMock.reset();
+  }
+}
--- a/application/backend/tests/test-elsa-settings.common.ts
+++ b/application/backend/tests/test-elsa-settings.common.ts
@@ -60,7 +60,7 @@ export const createTestElsaSettings: () => ElsaSettings = () => ({
     },
   ],
   logger: {
-    level: "warn",
+    level: "trace",
     transport: {
       targets: [
         {
--- a/application/backend/tests/test-elsa-settings.common.ts
+++ b/application/backend/tests/test-elsa-settings.common.ts
@@ -60,7 +60,7 @@ export const createTestElsaSettings: () => ElsaSettings = () => ({
     },
   ],
   logger: {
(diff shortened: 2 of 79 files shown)


andrewpatto:
Does the esbuild-jest surface errors like this earlier?

andrewpatto:
Whilst you are doing this - can you also fix the names of some of the services (global refactor/rename)?

In particular - we hadn't decided on plural v singular when first starting. But now should be consistently singular

but UsersService -> UserService
JobsService -> JobService

Doesn't need all the actual instances like usersService to be renamed (we can do that gradually).. but mind as well bite to bullet on the service type name now given how wide ranging the PR is..

mmalenic:
> Does the esbuild-jest surface errors like this earlier?

My hope is yes. Specifically for the dependency injection errors, it does seem to catch this in the tests. I think having the tests be transformed with esbuild will help in the future to avoid any esbuild issues.

**Note** that `ts-jest` is transitioning to using esbuild as a transformer as well, although I don't think this feature is ready yet. We could have two different kinds of jest configs; one using `esbuild-jest` and another using `ts-jest`. Once `ts-jest` fully supports esbuild we can remove `esbuild-jest`?

andrewpatto:
Happy with moving off ts-jest and just using esbuild-jest

application/backend/tests/service-tests/client-mocks.ts
```diff
@@ -0,0 +1,19 @@
...
+  private readonly stsClientMock = mockClient(STSClient);
+
+  public enable() {
+    this.stsClientMock.on(GetCallerIdentityCommand).resolves({});
+  }
+
+  public reset() {
+    this.stsClientMock.reset();
+  }
+}
```
mmalenic, line 19: This is partially on the way to solving #271. Sorry @DoxasticFox if you have already started on that.

application/backend/tests/test-elsa-settings.common.ts
```diff
@@ -60,7 +60,7 @@ export const createTestElsaSettings: () => ElsaSettings = () => ({
     },
   ],
   logger: {
-    level: "trace",
+    level: "warn",
```
andrewpatto: ??

andrewpatto (approved):

application/backend/tests/test-elsa-settings.common.ts
mmalenic: Ahh it was too noisy so I reduced it, but then I realised I could just pass in `--silent`.

application/backend/tests/service-tests/client-mocks.ts
DoxasticFox: Thanks for pointing this out, but no I haven't started, so all good :+1:

DoxasticFox (approved):

PR #301, opened 2023-04-21
Removed Convict

### LINKED ISSUE #300: Extend the configuration syntax to allow setting of more complex array values etc
Allow config settings with keys like

"nested.dac['rems'].password": "abcd"

Thinking of using a general JSONpath syntax in the keys - even though that is probably too powerful for our needs..

andrewpatto:
Merging the Settings type with the Configuration type


andrewpatto:
will quite possibly find the odd problem caused by this - but is hard to detect without just merging and getting on with using. (for instance there may be deployment in AWS issues)

DoxasticFox (approved):

PR #303, opened 2023-04-26
Patch FE 

williamputraintan:
patch #297 

- Fix incorrect `if-else` conditions
- Fix path that contain `.` (dot)

### DIFF
--- a/application/frontend/src/components/dataset/dataset-table.tsx
+++ b/application/frontend/src/components/dataset/dataset-table.tsx
@@ -123,7 +123,9 @@ export const DatasetTable: React.FC<{ includeDeletedFile: boolean }> = ({
                   <button
                     className={classNames("btn-table-action-navigate")}
                     onClick={async () => {
-                      navigate(encodeURIComponent(row.uri));
+                      navigate(
+                        encodeURIComponent(row.uri.replaceAll(".", "[dot]"))
+                      );
                     }}
                   >
                     view
--- a/application/backend/jest.config-service-tests-release.config.js
+++ b/application/backend/jest.config-service-tests-release.config.js
@@ -1,6 +1,8 @@
 module.exports = {
-  preset: "ts-jest",
   setupFiles: ["dotenv/config"],
   setupFilesAfterEnv: ["<rootDir>/tests/jest.setup.ts"],
   testMatch: ["<rootDir>/tests/service-tests/**/release*.test.ts"],
+  transform: {
+    "^.+\\.tsx?$": "esbuild-jest",
+  },
 };
--- a/application/backend/jest.db-tests.config.js
+++ b/application/backend/jest.db-tests.config.js
@@ -1,5 +1,7 @@
 module.exports = {
-  preset: "ts-jest",
   setupFilesAfterEnv: ["<rootDir>/tests/jest.setup.ts"],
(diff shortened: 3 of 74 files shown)


application/frontend/src/components/dataset/dataset-table.tsx
```diff
@@ -123,7 +123,9 @@ export const DatasetTable: React.FC<{ includeDeletedFile: boolean }> = ({
                   <button
                     className={classNames("btn-table-action-navigate")}
                     onClick={async () => {
-                      navigate(encodeURIComponent(row.uri));
+                      navigate(
+                        encodeURIComponent(row.uri.replaceAll(".", "[dot]"))
+                      );
                     }}
```
williamputraintan, line 129: Since databaseUUID is being replaced by datasetURI, it seems that path contain a dot sign (which URI usually has) might be a problem here. It keeps on returning 404 and the suggested current solution is to use [`vite-plugin-rewrite-all`](https://www.npmjs.com/package/vite-plugin-rewrite-all) which doesn't seems to work (if I have used it correctly). So I decided to convert `.` to `[dot]` before encoding it and get the URI back in reverse order when querying the BE. I think this is a reasonable approach (unless I am missing something)?

Related issue: `https://github.com/vitejs/vite/issues/2415`

DoxasticFox (approved):
I have only a very superficial understanding of what's going on, but just from reading the lines changed in this PR, the original code feels less wrong to me; the user _should_ require both of the relevant permissions to access something in the UI, right? Would it be better to change the logic for granting those permissions (i.e. `ALLOWED_DATASET_UPDATE` and `ALLOWED_OVERALL_ADMIN_VIEW`) so that they're both granted at the same time when appropriate, rather than relax the restrictions on the frontend? (I'm not politely suggesting this btw; I'm literally asking :joy:. Feel free to merge if you're confident of your changes.)

williamputraintan:
Yup, I think the idea is people who has the authority to update the dataset should be able to view the dataset as they need to select which dataset to update and also to see existing dataset if an update is required.

To some extend I do agree that they should both be granted at the same time when appropriate, but I suppose the `ALLOWED_OVERALL_ADMIN_VIEW` will able to view beyond just the dataset records which include all existing releases. But, yeah happy to make for the changes if people are in favour to this.

application/frontend/src/components/dataset/dataset-table.tsx
andrewpatto: At some point a long time ago I did add this into the notfound handler

```
if (request.url.includes(".")) {
          reply
            .code(404)
            .type("application/problem+json")
            .send({
              type: "about:blank",
              title: "Not Found",
              status: 404,
              detail: `File ${request.url} does not exist`,
            });

          return;
        }
```

that I guess might be playing a part (?) - though I would think this would only happen if someone does a Browser refresh. Regular navigation I would think would work?

andrewpatto (approved):

application/frontend/src/components/dataset/dataset-table.tsx
williamputraintan: Yes regular navigation works, just that refresh or copying and paste to the address bar doesn't.

Ah this might be the reason why the URL don't work when trying with the plugins, but I think we could leave this way for now.

PR #305, opened 2023-04-27
Split Manifest and Object Signing back out to two different UI boxes

### LINKED ISSUE #291: Split Manifest and Object Signing back out to two different UI boxes
We merged object signing and the basic manifest download at some point as they are both "downloading a CSV" - just with different columns.

However, I'd like to split them back out in the UI.

- Manifest is the only guaranteed always available download, that cannot be disabled by the custodian.
- It shouldn't download into a password protected zip - just should generate a TSV
- The endpoint for manifest might end up as a more public REST route
- It must never release actual data

"Object signing" is the more complex operation that gives acccess to data files - but is
- able to be disabled as a data sharing method
- will come down as an encrypted zip
- probably only ever an internal route

In the backend/route bodies - I'm happy for all the CSV/TSV processing to share common code - this is just about the actual route endpoints and frontend UI.

DoxasticFox:
Closes https://github.com/umccr/elsa-data/issues/291.

### DIFF
--- a/application/backend/src/api/routes/internal/release-routes.ts
+++ b/application/backend/src/api/routes/internal/release-routes.ts
@@ -519,28 +519,77 @@ export const releaseRoutes = async (
   fastify.post<{
     Body?: ReleasePresignRequestType;
     Params: { rid: string };
-  }>("/releases/:rid/tsv-manifest", {}, async function (request, reply) {
-    const { authenticatedUser } = authenticatedRouteOnEntryHelper(request);
+  }>(
+    "/releases/:rid/tsv-manifest-archive",
+    {},
+    async function (request, reply) {
+      const { authenticatedUser } = authenticatedRouteOnEntryHelper(request);
 
-    const releaseKey = request.params.rid;
+      const presignHeader = request.body?.presignHeader ?? [];
 
-    const manifest = await manifestService.getArchivedActiveTsvManifest(
-      presignedUrlService,
-      authenticatedUser,
-      releaseKey,
-      request.body?.presignHeader ?? []
-    );
+      const releaseKey = request.params.rid;
 
-    if (!manifest) {
-      reply.status(404).send();
-      return;
+      const manifest = await manifestService.getActiveTsvManifestAsArchive(
+        presignedUrlService,
+        authenticatedUser,
+        releaseKey,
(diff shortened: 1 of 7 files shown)


application/backend/src/api/routes/internal/release-routes.ts
```diff
@@ -519,28 +519,77 @@ export const releaseRoutes = async (
   fastify.post<{
     Body?: ReleasePresignRequestType;
     Params: { rid: string };
-  }>("/releases/:rid/tsv-manifest", {}, async function (request, reply) {
```
DoxasticFox, line 522: I split this into two routes:


* `/releases/:rid/tsv-manifest-plaintext` - Replies with TSV plaintext
* `/releases/:rid/tsv-manifest-archive` - Replies with TSV plaintext that's been put into a TSV archive. The TSV file given by this endpoint is the same as the other, except it can't contain the `objectStoreSigned` column.

application/backend/src/api/routes/internal/release-routes.ts
```diff
@@ -519,28 +519,77 @@ export const releaseRoutes = async (
...
+    Params: { rid: string };
+  }>(
+    "/releases/:rid/tsv-manifest-plaintext",
+    {},
+    async function (request, reply) {
+      const { authenticatedUser } = authenticatedRouteOnEntryHelper(request);
 
-    manifest.pipe(reply.raw);
-  });
+      const presignHeader = request.body?.presignHeader ?? [];
```
DoxasticFox, line 562: The changes in `application/common/elsa-types/schemas.ts` are so that TypeScript can figure out the type of `.presignHeader`.

application/frontend/src/components/access-box.tsx
```diff
@@ -0,0 +1,21 @@
...
+      defaultChecked={true}
+      name="presignHeader"
+      id={`chx-${field}`}
+      value={field}
+    />
+    <label className="uppercase" htmlFor={`chx-${field}`}>
+      {field}
+    </label>
+  </div>
+);
```
DoxasticFox, line 21: Copied from `manifest-form.tsx` and `object-signing-form.tsx`

andrewpatto (approved):

PR #306, opened 2023-04-28
Fix dataset e2e timeout

williamputraintan:
- Fix on broken e2e testing
- Add some e2e tests


andrewpatto:
Did you get the underyling causes of the working/not working randomness?

andrewpatto (approved):

williamputraintan:
> Did you get the underyling causes of the working/not working randomness?

Yes, its on the results order of edgeql query. When creating the release manually, it has the option to select the datasetUri from the given list and the `/dataset` API is responsible for it. The dataset API paginate so it never came up for the playwright to select from.

I added an URI sorting on the queries (so it could select for this test), and in real scenario I also added "load more dataset URI" button when needed.

PR #307, opened 2023-04-28
feat: editing alert

mmalenic:
### Changes
* Add an alert when a release has recently been edited - within the last 10 minutes.
    * Scrolls the alert into view when it occurs.
    * Animates similar to the DaisyUI button to grab attention.
* Add a property in the release schema to keep track of the user which last changed the release.

### UI
* I think using an alert here fits a bit nicer with the UI over a toast.
image

* This also could be a good way to render errors - which might be neater than showing them inside the boxes.


application/backend/dbschema/migrations/00002.edgeql
```diff
@@ -0,0 +1,9 @@
+CREATE MIGRATION m1f6twc6b27e6wkvvu3uckminarlcrtr5ok7qke3fop5n7onf4vora
+    ONTO m1lt44twvfd75hhuuyjgxhwxzsm2uteenmn3gnus2nuqztoxqvsgla
+{
+  ALTER TYPE release::Release {
+      CREATE REQUIRED PROPERTY lastUpdatedSubjectId -> std::str {
+          SET REQUIRED USING ('unknown');
```
mmalenic: Might be good to reset the database to no migrations so that this doesn't need to be specified.

andrewpatto:
Looks good - happy for any UI component - I only referred to "toast" as the kind of interaction I was envisaging - this looks like it hits the spot.

application/backend/src/business/db/release-queries.ts
```diff
@@ -1,17 +1,20 @@
...
+    subjectId: e.str,
   },
   (params) =>
     e.update(e.release.Release, (r) => ({
       filter: e.op(r.releaseKey, "=", params.releaseKey),
-      set: { lastUpdated: e.datetime_current() },
+      set: {
+        lastUpdated: e.datetime_current(),
+        lastUpdatedSubjectId: params.subjectId,
+      },
```
andrewpatto: can I be annoying and get you to rewrite this in edgeql? It seems like a nice easy one - and I'd definitely prefer us to move to the all edgeql pattern eventually.. (one PR at a time!)

application/frontend/src/pages/releases/releases-master-page.tsx
```diff
@@ -24,6 +28,9 @@ import { trpc } from "../../helpers/trpc";
  */
 export const ReleasesMasterPage: React.FC = () => {
   const REFRESH_JOB_STATUS_MS = 5000;
+  const ALERT_RELEASE_EDITED_TIME: Millisecond = 600000;
```
andrewpatto, line 31: not super worried about this - but thinking it might be the type of config that gets set in the backend config and passed down via the App html variables?

andrewpatto (approved):

andrewpatto:
What is the behaviour of the close button on the alert? Does it then come back the next page they navigate to?

application/backend/dbschema/migrations/00002.edgeql
andrewpatto: happy if you do the reset in this PR..

application/frontend/src/pages/releases/releases-master-page.tsx
mmalenic: I feel like this is more of a user-facing config, because it's all calculated in the front-end anyway. We could allow it to be configured in the account config drop-down? Or otherwise, I'm happy to add this to the backend?

application/frontend/src/components/alert.tsx
```diff
@@ -0,0 +1,116 @@
...
+      enter="transition-opacity duration-75"
+      enterFrom="opacity-0"
+      enterTo="opacity-100"
+      leave="transition-opacity duration-150"
+      leaveFrom="opacity-100"
+      leaveTo="opacity-0"
+      ref={alertRef}
+    >
+      <div>
+        <svg
```
williamputraintan, line 80: Just an alternative for a simpler Icon code

```
<FontAwesomeIcon icon={faTriangleExclamation} />
```

application/frontend/src/components/alert.tsx
```diff
@@ -0,0 +1,116 @@
...
+        </svg>
+        <span>{description}</span>
+      </div>
+      <button
+        className="btn-ghost btn-sm btn"
+        onClick={() => {
+          setDismissed(true);
+        }}
+      >
+        <svg
```
williamputraintan: Ditto
```
<FontAwesomeIcon icon={faX} />
```

williamputraintan (approved):
LGTM

DoxasticFox (approved):

application/backend/src/business/db/release-queries.ts
mmalenic: Sure, good idea.

application/frontend/src/components/alert.tsx
mmalenic: Thanks! I think for the `faTriangleExclamation` I'll keep the svg version because the regular version of the triangle is not a free icon.

mmalenic:
> What is the behaviour of the close button on the alert? Does it then come back the next page they navigate to?

When it was closed it would never come back, even when the `lastUpdated` properties changed. I've fixed that now so that it comes back only if the `lastUpdated` properties are different (not dependent on what page the user is on).

PR #308, opened 2023-05-02
Turns out jsonpath can't be bundled - so need to manually deploy its …

andrewpatto:
…dependency tree

### DIFF
--- a/application/Dockerfile
+++ b/application/Dockerfile
@@ -136,9 +136,16 @@ COPY --from=edgedb-builder /build/dbschema/migrations/ /a/dbschema/migrations/
 COPY --from=app-builder /build/backend/config/ /a/config/
 COPY --from=app-builder /build/backend/locales/ /a/locales/
 
-# bring in any node_modules incompatible with bundling
+# bring in any node_modules incompatible with bundling (including dependencies of these modules too)
 COPY --from=app-builder /build/backend/node_modules/node-gyp-build/ /a/node_modules/node-gyp-build/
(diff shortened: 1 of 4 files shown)


mmalenic (approved):
Ahh, I was just looking into this one.

How would you feel about adding something like this to the github actions `tests.yml`?

```yaml
# esbuild should succeed and contain no warnings
- name: esbuild check
  run: |
    output=$(npx node bundle-build.js)
    if [[ $? == 0 ]]; then
      echo "esbuild ran successfully"
    else
      echo "esbuild did not succeed"
      echo "$output"
      exit 1
    fi
  working-directory: application/backend
```

This would check that `bundle-build.js` succeeds and contains no output.

andrewpatto:
> Ahh, I was just looking into this one.
> 
> How would you feel about adding something like this to the github actions `tests.yml`?
> 
> ```yaml
> # esbuild should succeed and contain no warnings
> - name: esbuild check
>   run: |
>     output=$(npx node bundle-build.js)
>     if [[ $? == 0 ]]; then
>       echo "esbuild ran successfully"
>     else
>       echo "esbuild did not succeed"
>       echo "$output"
>       exit 1
>     fi
>   working-directory: application/backend
> ```
> 
> This would check that `bundle-build.js` succeeds and contains no output.

Yes - that's a good idea and a quick win - so happy for that to be added.

There are other ways the resulting Docker build can fail though - I think at some point in the literal Docker building action we will need to check that.

PR #309, opened 2023-05-02
Temporarily skip broken service tests and add the rest to CI

DoxasticFox:
Relates to https://github.com/umccr/elsa-data/issues/271.

A couple of questions:

1. @andrewpatto - The tests in `application/backend/tests/service-tests` are what you were referring to, right? They mostly seem to be running already. The mocking just needs some tweaking for the tests I've skipped in this PR. I don't think we'll need to move to something as heavy as localstack for now; fixing the mocks should be fine. All the release tests seem fine too.
2. Is everyone okay with me skipping these tests unconditionally, or should I only skip them in CI? I'll raise a couple of follow-up PRs to fix the mocks shortly. I just figured adding the other service tests to the build was a quick win that'll stop further inadvertent breakages.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -94,6 +94,10 @@ jobs:
         run: npm run test:integration
         working-directory: application/backend
 
+      - name: Run service tests
+        run: npm run test:service
+        working-directory: application/backend
+
       # setup a cache explicitly for Playwright dependencies (browser dependencies at the OS level)
       # https://justin.poehnelt.com/posts/caching-playwright-in-github-actions/
       - uses: actions/cache@v3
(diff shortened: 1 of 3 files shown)


andrewpatto:
Yes those are the right tests.

Happy for you to use your judgement on what should be quick fixed v skipped. I think assigning out follow up PRs is a good idea.

I just wanted to get them to a stable state - didn't want you to feel you had to fix all the tests themselves as well in this PR.

andrewpatto:
I'm sure I have more than that fail when I run them in Webstorm??

Have you done any jest launches with _no_ AWS credentials/config at all? (that is how my webstorm runs them)

andrewpatto:
I mean I guess the fact it is passing in github actions means that they do work with no AWS

andrewpatto (approved):

PR #310, opened 2023-05-02
ci: add esbuild ci step

mmalenic:
Adds an esbuild step to the tests github action.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -80,7 +80,7 @@ jobs:
       # esbuild should succeed and contain no warnings
       - name: esbuild check
         run: |
-          output=$(npx node esbuild 2>&1)
+          output=$(npx node bundle-build.js 2>&1)
           if [[ $? == 0 ]]; then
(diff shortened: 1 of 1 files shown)


andrewpatto (approved):

.github/workflows/tests.yml
```diff
@@ -77,6 +77,19 @@ jobs:
         run: npx tsc --noEmit
         working-directory: application/backend
 
+      # esbuild should succeed and contain no warnings
+      - name: esbuild check
+        run: |
+          output=$(npx node bundle-build.js 2>&1)
+          if [[ $? == 0 ]]; then
+            echo "esbuild ran successfully"
```
andrewpatto, line 86: Is there any problems with the generated esbuild files from this step? Like might they accidentally end included in the docker build files or something? Or affect the e2e tests accidentally?

mmalenic: I don't think it would affect them because they should be building their own code anyway, although I've added an `rm -r server` after checking esbuild just in case.

PR #312, opened 2023-05-02
Re-enable service tests skipped in #309

### LINKED ISSUE #271: Rescue the service tests including splitting out to cloud v non-cloud

DoxasticFox:
Closes https://github.com/umccr/elsa-data/issues/271.

### DIFF
--- a/application/backend/babel.config.js
+++ b/application/backend/babel.config.js
@@ -1,0 +1,6 @@
+module.exports = {
+  presets: [
+    ["@babel/preset-env", { targets: { node: "current" } }],
+    "@babel/preset-typescript",
+  ],
+};
--- a/application/backend/package-lock.json
+++ b/application/backend/package-lock.json
@@ -65,7 +65,7 @@
         "humanize-plus": "1.8.2",
         "i18n": "0.15.1",
         "json5": "2.2.3",
-        "jsonpath": "^1.1.1",
+        "jsonpath": "1.1.1",
         "jszip": "3.10.1",
         "lodash": "4.17.21",
         "maxmind": "^4.3.8",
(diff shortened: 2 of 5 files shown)


application/backend/babel.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  presets: [
+    ["@babel/preset-env", { targets: { node: "current" } }],
+    "@babel/preset-typescript",
+  ],
+};
```
mmalenic, line 6: Is there a reason babel is being used here?

mmalenic (approved):

application/backend/babel.config.js
DoxasticFox: When I added `jest.mock("../../src/business/services/aws/aws-helper",...` to `ag-s3-import.test.ts`, the tests started failing, complaining about the type annotation on `let edgedbClient: edgedb.Client`. The error message told me to do [this](https://jestjs.io/docs/getting-started#using-typescript).

mmalenic: Ahh interesting. I [previously removed ](https://github.com/umccr/elsa-data/pull/299/files#diff-fbb873173b8c84064b684026f55505663331f946a018c48a77a0ea2a3f3c7d0b)ts-jest in favour of using [esbuild with jest](https://www.npmjs.com/package/esbuild-jest). Not sure if that impacts this.

mmalenic: Not sure if this issue is related: https://github.com/aelbore/esbuild-jest/issues/57

DoxasticFox: Possibly. It's the same error. https://gist.github.com/DoxasticFox/887f3e5e872cbf7a71c966b966d54ec1

williamputraintan (approved):

application/backend/babel.config.js
DoxasticFox: I'm going to merge anyway even though I'm not satisfied with what I've done in this PR, because I don't know how to improve it and it's still better to run the tests than not to, I guess.

PR #313, opened 2023-05-03
refactor(backend): try non-strict parsing when strict parsing fails

mmalenic:
Removes strict parsing to avoid errors when unknown keys are found by providers.


andrewpatto (approved):

andrewpatto:
I think this is good for now - but I would possibly like 'production' deployments to still retain the hard fail. But we can deal with that when we get closer to actual production.

PR #314, opened 2023-05-03
fix(backend): add transport targets to deploy config

mmalenic:
Adds a transport targets field to the deployed logger config to avoid errors with invalid types

### DIFF
--- a/application/backend/config/dev-deployed.json5
+++ b/application/backend/config/dev-deployed.json5
@@ -3,6 +3,12 @@
   // but we want to get all the logs so level of debug
   logger: {
     level: "debug",
+    transportTargets: [
+      {
+        target: "pino-pretty",
+        options: { destination: 1, minimumLevel: "debug" },
+      },
+    ],
   },
   htsget: {
     url: "https://htsget.elsa.dev.umccr.org",
--- a/application/backend/config/dev-deployed.json5
+++ b/application/backend/config/dev-deployed.json5
@@ -3,12 +3,6 @@
   // but we want to get all the logs so level of debug
   logger: {
     level: "debug",
-    transportTargets: [
-      {
-        target: "pino-pretty",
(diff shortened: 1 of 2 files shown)


andrewpatto:
So we need to start thinking about the distinction between configs for dev itself, configs for running locally but in the dev account, and configs for actually deploying an image into dev.
So everything in the config/ folder won't make it through into the actual Docker imag..

andrewpatto:
Configs for actually dev deployment - should live in the assets folder in the aws-deploy project

andrewpatto:
(not saying this PR is wrong - just saying I don't know where we currently are in all of this)

mmalenic:
Yes, you're right. Maybe there shouldn't even be a dev-deployed config in this repository?

andrewpatto:
Yes - I think it pre-dates us even having a deployable AWS stack.

But we do have a use case for _running_ Elsa locally - but with dev AWS credentials - and so I think some of these settings inhabited both worlds?

andrewpatto:
For instance - the pretty printer for debug should *only* be enabled for local dev runs (it puts colours into the console logs).

For cloudwatch the colours will be useless - so we definitely don't want this for AWS actual deployments

mmalenic:
Yes that's true. I guess it can help with calling `file('dev-deployed')` in the meta configs of the aws-deploy repo, so it does sort of serve two purposes. Although, probably any config for deployment should just be in that repo instead.

andrewpatto:
Yes -lets split them - anything in config/ in the Elsa repo should *only* make sense for local dev deployment

Anything needed by the dev deployment should go over into the other aws-deploy repo

mmalenic:
Okay, I think this PR is now correct, as it fixes the default behaviour for the logger config, which was throwing an error before.

andrewpatto (approved):

PR #315, opened 2023-05-03
Minor Release Page

williamputraintan:
- Convert `/release/{releaseKey}` route to TRPC
- Had a the externalIdentifier picker for manifest files
- Hide Activate/Deactivate button if user is not an "Admin" role

Screenshot 2023-05-03 at 17 52 21

### DIFF
--- a/application/frontend/src/pages/releases/releases-master-page.tsx
+++ b/application/frontend/src/pages/releases/releases-master-page.tsx
@@ -2,7 +2,7 @@ import React, { useEffect, useState } from "react";
 import { useQuery, useQueryClient } from "@tanstack/react-query";
 import { Outlet, useParams } from "react-router-dom";
 import { Box } from "../../components/boxes";
-import { REACT_QUERY_RELEASE_KEYS, specificReleaseQuery } from "./queries";
+import { makeReleaseTypeLocal } from "./queries";
 import { isUndefined } from "lodash";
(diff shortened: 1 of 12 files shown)


application/frontend/src/pages/releases/releases-master-page.tsx
```diff
@@ -46,12 +45,16 @@ export const ReleasesMasterPage: React.FC = () => {
...
+  const releaseQuery = trpc.releaseRouter.getSpecificRelease.useQuery(
+    { releaseKey },
+    {
+      onError: (error: any) => setError({ error, isSuccess: false }),
+      onSuccess: (_: any) => setError({ error: null, isSuccess: true }),
+    }
+  );
+  const localReleaseData = releaseQuery.data
+    ? makeReleaseTypeLocal(releaseQuery.data)
+    : null;
```
andrewpatto, line 57: is it possible to roll this into the events etc of useQuery()?

andrewpatto (approved):

application/frontend/src/pages/releases/releases-master-page.tsx
williamputraintan: I tried on `queryFn` for post-processing the data but I think TRPC override this to call the function to return the results.

One solution I think would use the onSuccess but we need to use our own states so I think would be better this way

PR #316, opened 2023-05-04
Improve Dataset Query Performance

williamputraintan:
- Splitting All Dataset query and Single Dataset Query
- Split the Large Single Dataset query into 2 smaller queries
- In local dev, time taken for the dataset table query should be pretty much instant, and the single dataset page should be similar to the release page.


application/backend/dbschema/queries/dataset/getDatasetSummary.edgeql
```diff
@@ -1,195 +0,0 @@
...
-            fastqArtifact.reverseFile.size,
-            vcfArtifact.vcfFile.size,
-            vcfArtifact.tbiFile.size,
-            bamArtifact.bamFile.size,
-            bamArtifact.baiFile.size,
-            cramArtifact.cramFile.size,
-            cramArtifact.craiFile.size,
-          }),
-
-          # List all type of artifacts
```
williamputraintan, line 139: Interesting to know that these are very time consuming

andrewpatto: Were you just doing stop-watch checks for speed - or do you have a way of instrumenting the edge queries to see how they perform?

andrewpatto:
Given we are rewriting these - can we start to get a common naming pattern on the paging queries.. ala getUsers

```
select {
  total := count(u),
  data := (
    select u {
      id,
      subjectId,
      email,
      displayName,
...
      rightsCount := (1 if .isAllowedRefreshDatasetIndex else 0) +
                     (1 if .isAllowedCreateRelease else 0) +
                     (1 if .isAllowedOverallAdministratorView else 0)
    }
    order by .isCaller DESC then .rightsCount DESC then .lastLoginDateTime DESC
    offset <optional int64>$offset
    limit  <optional int64>$limit
  )
}
```

So common names `total`, `data`, `offset` and `limit`

application/backend/dbschema/queries/dataset/getDatasetSummary.edgeql
williamputraintan: No, I just check manually the "stop-watch checks" approach.

andrewpatto:
Otherwise looks good.. I feel the need for speed..

application/backend/dbschema/queries/dataset/getDatasetSummary.edgeql
williamputraintan: It took ~4 seconds before to complete, and removing these takes out about 3 seconds out to finish the query in local.

andrewpatto (approved):

andrewpatto:
You'll always be my goose @williamputraintan

application/backend/dbschema/queries/dataset/getDatasetSummary.edgeql
andrewpatto: The Edge people have promised some sort of profiler/query analyser at some point - which will be useful to see which particular bits of a query are problematic

PR #318, opened 2023-05-05
Audit table fixes

andrewpatto:
Corrections to audit table to match other table formats (basically removing most formatting and letting daisy ui defaults)
Default filter for top level audit page is "user" only
Rolled duration column underneath time (duration is rarely filled in and is unlikely for sorting/filtering)
(temp) Removed possibility of going to detail page (need to work out how it works with breadcrumbs etc)

### DIFF
--- a/application/backend/src/api/api-auth-routes.ts
+++ b/application/backend/src/api/api-auth-routes.ts
@@ -168,7 +168,11 @@ export const callbackRoutes = async (
     const authUser = await userService.upsertUserForLogin(
       idClaims.sub,
       idClaims.name,
-      idClaims.email
+      idClaims.email,
+      {
+        ip: request.ip,
+        hostname: request.hostname,
+      }
     );
 
     if (!authUser) {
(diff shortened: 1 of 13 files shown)


andrewpatto:
Screenshot 2023-05-05 at 2 55 53 pm

andrewpatto:
Changed the language throughout to refer to only Audit Events and Audit Event

application/backend/src/api/api-auth-routes.ts
```diff
@@ -168,7 +168,11 @@ export const callbackRoutes = async (
     const authUser = await userService.upsertUserForLogin(
       idClaims.sub,
       idClaims.name,
-      idClaims.email
+      idClaims.email,
+      {
+        ip: request.ip,
+        hostname: request.hostname,
+      }
```
andrewpatto, line 175: To better have examples of audit entries with details - we mind as well always log extra details of logins

williamputraintan (approved):

mmalenic (approved):
Looks good! I don't think anything has been broken.

PR #323, opened 2023-05-11
feat(backend): use EdgeDb 3.0 in Dockerfile

mmalenic:
* Use EdgeDb 3.0 when building queries.


andrewpatto (approved):

andrewpatto:
There's some other spots also? (not in the Dockerfile - but elsehwere??)

mmalenic:
> There's some other spots also? (not in the Dockerfile - but elsehwere??)

Ahh yes, I've added the nightly server and cli to the tests. I've also added it to the `edgedb.toml`, is everyone okay with that? If you use `edgedb project init`, you might need to update to the nightly cli and start with `edgedb project init --server-version=3.0-beta.1`. See [here](https://www.edgedb.com/docs/changelog/3_x#). /cc @williamputraintan @DoxasticFox 

Is that all the spots?

PR #325, opened 2023-05-11
Improve permissions handling for audit entries

### LINKED ISSUE #324: Users should be able to see the details of their own audit entries
The permissions enforced on seeing the details of a log entry are too strict - I should be able to see my own entries (I think?)

andrewpatto:
Renamed audit log service to audit entry service
Added an EdgeDb query that knows the rules about audit entries Added loading divs to audit table
Started on db tests


andrewpatto:
Ok - so this is a much larger function - that has within it all the rules for audit logs. So preferably we move to this eventually.

HOWEVER - I can't work out how to do parametrised EdgeQl ordering - which means this query is less functional than what we have. So I've only put this into the main code path for 'fetch audit details'.

I will make a new issue to revisit how EdgeQl can do sorting - and that ticket can handle replacing some of the existing queries with this one.

andrewpatto:
Added some wait spinning cursors to the audit event table front end as well

application/backend/dbschema/queries/audit-event/auditEventGetSomeByUser.edgeql
```diff
@@ -0,0 +1,93 @@
...
+# truncate the amount of JSON details data returned
+#
+# This query can filter by audit event type or can be used to return a single audit event
+#
+# This query applies all required permissions rules to ensure that the given
+# user is allowed to see the events and returns an empty result if not
+#
+# WIP - this cannot filter/order the resulting set with as much power as the
+# hand written Typescript queries - so this is not yet enabled for most querying in Elsa Data
+
```
andrewpatto, line 11: This is the main unification of other typescript queries - main advantage being the visibility rules can be put all in one spot

application/backend/tests/db-tests/edgedb-audit-entry-queries.test.ts
```diff
@@ -0,0 +1,346 @@
...
+const SUBJECT_EMAIL_2 = "subject@2.com";
+
+const SUBJECT_ID_3 = "subjectid3";
+const SUBJECT_DISPLAY_NAME_3 = "Subject 3";
+const SUBJECT_EMAIL_3 = "subject@3.com";
+
+const testContainer = registerTypes();
+
+describe("edgedb audit entry tests", () => {
+  let edgeDbClient: Client;
```
andrewpatto, line 33: a unified query for audit entries means I can also unify the testing and be a bit more thorough at the lower db test level

application/backend/dbschema/queries/audit-event/auditEventGetSomeByUser.edgeql
```diff
@@ -0,0 +1,93 @@
...
+      isUserAuditEvent,
+      whoSubjectId,
+      whoDisplayName,
+      isReleaseAuditEvent,
+      releaseId := .release.id,
+      detailsAsPrettyString := to_str(.details, 'pretty')[0 : m] if exists(.details) else "",
+      detailsWereTruncated := exists(.details) and m >= 0 and len(to_str(.details, 'pretty')) >= m
+    }
+    # WIP need to add some dynamism here to allow order fields
+    order by .occurredDateTime desc
```
mmalenic, line 90: Previously I had all the possible order by expressions listed out in a [chained if else](https://github.com/umccr/elsa-data/blob/b20b80979fa37e09d2ded59c3d3c15a2b2cf8c59/application/backend/src/business/db/audit-log-queries.ts#L95-L125). E.g. Something like `order by <json>.occurredDateTime if <str>$orderByProperty = "occurredDateTime" else ... desc`. This might be possible in EdgeQL too? But it's a bit clunky, which is why I changed it in the client library queries. Note, I think the if else needs to be a common type, maybe it could be cast to `str` or `json`?

Maybe there's a neat way to do this with some sort of set operation?

application/frontend/src/components/audit-event/details-row.tsx
```diff
@@ -55,24 +58,35 @@ export const DetailsRow = ({ objectId }: DetailsRowProps): JSX.Element => {
...
+  if (detailsQuery.isError)
+    return (
+      <EagerErrorBoundary
+        message={"Something went wrong displaying audit event details."}
+        error={detailsQuery.error}
+        styling={"bg-red-100"}
+      />
+    );
+
+  return <></>;
```
mmalenic, line 91: Nice!

mmalenic (approved):
Looks good to me.

application/backend/dbschema/queries/audit-event/auditEventGetSomeByUser.edgeql
andrewpatto: Yeah I could work out the chaining (albeit clunky) - but couldn't work out chaining with ASC and DESC?

andrewpatto: As in, I couldn't make the ASC or DESC be parameterised..

andrewpatto: (I don't so much mind the casting to str)

mmalenic: Yeah.. I can't see a way to do that either. Other than having two of the same query.

PR #329, opened 2023-05-15
Ability to load our 10G dataset more realistically

### LINKED ISSUE #327: Make a 10G dataset creator that can work outside of AWS
Whilst we have real actual AWS files in UMCCR for the 10G dataset - it would be useful for demo purposes if it could also exist as a fake mock only dataset for those without access to the UMCCR buckets.

So this is to investigate what we can do in terms of mocking this..

andrewpatto:
Mocks a filename structure so that loading the 10G dataset goes via the main AG loader


andrewpatto:
Because this does not hit AWS - it can be used in *both* E2E/local testing - but also represent real datasets in AWS if we want.

andrewpatto:
Have started on some level of "loader" types for datasets - though we need to revisit. Works for the moment without breaking the substantial amount of test code etc

application/backend/src/api/routes/internal-trpc/dataset-router.ts
```diff
@@ -52,9 +52,30 @@ export const datasetRouter = router({
...
+            loader
+          );
+          break;
+        case "australian-genomics-directories-demo":
+          await ctx.agS3IndexService.syncWithDatabaseFromDatasetUri(
+            datasetUri,
+            user,
+            loader
+          );
+          break;
```
andrewpatto, line 75: this is the pattern I would like to move to - but we need to refactor `syncWithDatabaseFromDatasetUri` so that it can accept a more abstract definition of the storage it is running over (so it can run over 'fake' as well as real clouds).

So `syncWithDatabaseFromDatasetUri` currently has a switch inside it too that detects the demo cases..

application/backend/src/business/db/dataset-queries.ts
```diff
@@ -27,6 +27,6 @@ export const selectDatasetIdByDatasetUri = (datasetUri: string) =>
   e
     .select(e.dataset.Dataset, (d) => ({
       id: true,
-      filter: e.op(d.uri, "ilike", datasetUri),
+      filter: e.op(d.uri, "=", datasetUri),
```
andrewpatto, line 30: this should never have been ilike. Dataset uris are exact matches

application/backend/src/business/services/australian-genomics/s3-index-import-service.ts
```diff
@@ -715,19 +734,66 @@ export class S3IndexApplicationService {
...
+              async (url: string) => {
+                if (!(url in phases[1].fileContent))
+                  throw new Error(
+                    `Demo dataset loader missing manifest content at ${url}`
+                  );
+
+                return phases[1].fileContent[url];
+              }
+            );
+        } else throw Error("AG demo loader used with unknown dataset URI");
```
andrewpatto, line 792: So this lives outside the AG index service - but requires a major refactor

DoxasticFox (approved):

application/backend/src/business/db/dataset-queries.ts
williamputraintan: Agree! Sorry my bad

williamputraintan (approved):

application/backend/src/business/services/australian-genomics/s3-index-import-service.ts
```diff
@@ -715,19 +734,66 @@ export class S3IndexApplicationService {
...
+          await this.datasetService.getDemonstrationStoragePrefixFromDatasetUri(
+            datasetUri
+          );
+
+        if (!fakeS3UrlPrefix) {
+          console.warn("No demonstration storage prefix");
+          return;
+        }
+
+        if (datasetUri === TENG_URI) {
```
mmalenic, line 773: Maybe not necessary for this PR, but would it be good to have a mechanism that's more generic and can load any demo dataset?

mmalenic (approved):

application/backend/src/business/services/australian-genomics/s3-index-import-service.ts
andrewpatto: Yes - this is temporary until we can move this logic out into a more general service (and at the same time would be good to establish a more generic mechanism)

PR #331, opened 2023-05-15
FrontEnd

### LINKED ISSUE #272: `alert()` is called repeatedly for same error
On an Elsa instance without AWS properly configured, if you navigate to the "AWS S3 VPC SHARE" tab in the "Access Data" pane, you'll see an alert pop up which keeps reappearing after being dismissed. The error appearing is expected; It's telling me about a legitimate problem with my configuration. The problem is specifically that the alert keeps reappearing after being dismissed, until you navigate to a different tab (e.g. "MANIFEST").

I suspect the main issue is that alerts are happening [here](https://github.com/umccr/elsa-data/blob/60b811d6896a5ca6a447f2cfbb5d84c1d13e668a/application/frontend/src/providers/trpc-provider.tsx#L28) when they should instead happen in a `useEffect` hook.

I think the alerts might also happen with less frequency if we tried harder to avoid unnecessary re-renders. ([Related thread](https://github.com/umccr/elsa-data/pull/254#discussion_r1142779932)). I've attached a video of the issue below. You can see that each time the component is re-rendered, it produces (an) alert(s) and makes (a) HTTP request(s). The alerts and network activity associated with that component continue until you navigate to a different tab.

https://user-images.githubusercontent.com/11529449/228955715-cbcb8aa1-5294-4b3f-b6e9-9d89a72147a3.mov

### LINKED ISSUE #330: Frontend Responsiveness
Many UI component (mostly tables) is overlapping on smaller screen. I think we should sort this out for our MVP.

williamputraintan:
- UI Responsiveness 
- Use in-house alert
- Css fixes
- Remove "Edit mode" in user configuration

Fix #330 
Fix #272 

Screenshot 2023-05-15 at 17 54 22

<br/>

Screenshot 2023-05-15 at 18 00 51
 
Screenshot 2023-05-15 at 18 07 38


application/frontend/src/components/boxes.tsx
```diff
@@ -41,7 +41,7 @@ export const Box: React.FC<PropsWithChildren<BoxProps>> = ({
       )}
       <div
         className={classNames("card-body", {
-          sepia: applyIsLockedStyle,
+          "sepia-[.3]": applyIsLockedStyle,
```
andrewpatto: Nice - I wanted sepia - but that's a better amount of sepia!

application/frontend/src/components/data-egress/data-egress-detailed-table.tsx
```diff
@@ -44,43 +45,42 @@ export function DataEgressDetailedTable({
       <div className="mb-2	text-gray-500">
         A detailed events on every events egress from data storage.
       </div>
-      <table className="table-compact table w-full">
-        <thead>
+      <Table
```
andrewpatto, line 48: should we be adding classNames to this Table so it matches up with our Daisy table formatting (see releases table)

application/frontend/src/providers/logged-in-user-config-relay-provider.tsx
```diff
@@ -23,18 +24,22 @@ export type LoggedInUserConfigRelay = {
...
 ) => {
+  const loggedInUser = useLoggedInUser();
+
   const datasetsQuery = trpc.datasetRouter.getConfiguredDatasets.useQuery(
     undefined,
     {
       refetchOnWindowFocus: false,
       refetchOnReconnect: false,
       retry: false,
+      enabled: !!loggedInUser,
```
andrewpatto, line 35: good catch

andrewpatto (approved):

mmalenic (approved):
Nice! Is this Elsa mobile? ;) (Should we make an app - react-native, flutter?)

andrewpatto:
> Nice! Is this Elsa mobile? ;) (Should we make an app - react-native, flutter?)

I think we should start an Elsa hardware division making our own line of tablets and mobile devices

DoxasticFox (approved):

application/frontend/src/components/data-egress/data-egress-detailed-table.tsx
williamputraintan: Do you mean the `table` class from DaisyUI?

[The custom `<Table/>` component ](https://github.com/umccr/elsa-data/blob/william-patches/application/frontend/src/components/tables.tsx) by default has the `table` className from the DaisyUI

PR #332, opened 2023-05-16
Allow namespace to be specified in config

andrewpatto:


andrewpatto:
The AG deployments now have differing namespaces because they are in the same account - so I need a mechanism where this can be configured

williamputraintan (approved):

PR #335, opened 2023-05-16
Patch CSS FrontEnd

williamputraintan:
- Sticky footer is back
- Fix Consent PopUp (remove [Reactjs-popup](https://react-popup.elazizi.com/) and use [daisyUI-dropdown](https://daisyui.com/components/dropdown/) instead)

Screenshot 2023-05-16 at 19 45 14

### DIFF
--- a/application/frontend/src/index.css
+++ b/application/frontend/src/index.css
@@ -9,7 +9,7 @@ body {
 
 /* Major cards as displayed down the page */
 .card-major {
-  @apply card card-compact bg-base-100 shadow-lg;
+  @apply card card-compact static bg-base-100 shadow-lg;
 }
 
 .btn-standard-action-trigger-dialog {
--- a/application/frontend/src/providers/show-alert-provider.tsx
+++ b/application/frontend/src/providers/show-alert-provider.tsx
@@ -28,32 +28,31 @@ export default function ShowAlert(props: { children: React.ReactNode }) {
   return (
     <CtxProvider value={{ show: handleIsShowing }}>
       {isShowing && (
-        <div className="absolute z-20 flex w-full justify-center">
-          <div className="container alert mt-4 shadow-2xl">
-            <div>
+        <>
+          <div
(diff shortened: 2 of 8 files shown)


application/frontend/src/index.css
```diff
@@ -9,7 +9,7 @@ body {
 
 /* Major cards as displayed down the page */
 .card-major {
-  @apply card card-compact bg-base-100 shadow-lg;
+  @apply card card-compact static bg-base-100 shadow-lg;
```
andrewpatto, line 12: how does this alter the sticky footer?

andrewpatto: oh - wait - its the CtxProvider.. never mind..

application/frontend/src/providers/show-alert-provider.tsx
```diff
@@ -28,32 +28,31 @@ export default function ShowAlert(props: { children: React.ReactNode }) {
...
+            <div className="container alert mt-4 shadow-2xl">
               <div>
-                <h3 className="font-bold">{alertProps.title}</h3>
-                <div className="text-xs">{alertProps.description}</div>
+                <div>
+                  <h3 className="font-bold">{alertProps.title}</h3>
+                  <div className="text-xs">{alertProps.description}</div>
+                </div>
+              </div>
+              <div className="flex-none">
```
andrewpatto: can we daisy UI something here so it looks a little bit the same as the other alerts (I realise this one needs to sit over the top of the others so it isn't quite the same)

andrewpatto (approved):

application/frontend/src/providers/show-alert-provider.tsx
williamputraintan: Yup, this is the code from DaisyUI. Realising this, I changed it to re-use the alert component instead here.

application/frontend/src/index.css
williamputraintan: Yes, this is for the popup component.

PR #342, opened 2023-05-17
Create a mechanism for branding a deployed site

### LINKED ISSUE #311: Create a mechanism for branding a deployed site
Need a mechanism by which we can brand different deployments of Elsa. Obviously we can't build the logos etc into the actual base Docker image - so it needs to be a mechanism where a custom docker with elsa BASE (and some config) allows branding to be added.

Open for discussion in this ticket as to what can be branded.. 

Logos? Landing page text? Logged in landing page text?

DoxasticFox:
## Small login page

Screenshot_20230517_115439

## Large login page

Screenshot_20230517_115418

## Small logged in user

Screenshot_20230517_115518

## Large logged in user

Screenshot_20230517_115503


application/backend/src/app.ts
```diff
@@ -264,15 +276,32 @@ export class App {
...
+            return "&lt;";
+          case ">":
+            return "&gt;";
+          case "\\":
+            return "&#92;";
+          default:
+            return c;
+        }
+      });
+    };
```
andrewpatto, line 294: aren't the attribute values themselves eventually in HTML attribute quotes? Which means that we only need to escape things unsafe in string? backslash and quote?
Or are ampersands not even allowed in html attributes? (I can't remember)

andrewpatto:
Can we get a short markdown - in docs - I know we don't really have any structures yet so just anywhere - that explains what someone would need to do in their downstream Docker image in order to use this.

i.e. a little snippet of

```
ADD /my-logos/garvan.jpg
```

and then what the corresponding branding config settings would be.

application/backend/src/app.ts
```diff
@@ -227,6 +229,16 @@ export class App {
       }
     );
 
+    const logoUriRelative = this.settings.branding?.logoUriRelative;
+    const logoPath = this.settings.branding?.logoPath;
+    if (logoUriRelative && logoPath) {
+      await this.server.get(logoUriRelative, async (_, reply) => {
+        const mimeType = mime.lookup(logoPath);
```
andrewpatto, line 236: I think possibly fastify has a plugin/feature that can just "send" static files given a filename like logoPath
(and it will do the mime stuff itself)

andrewpatto (approved):

application/backend/src/app.ts
DoxasticFox: I think I erred in escaping backslashes, but the rest seems fine to me. w3's docs on HTML and XML syntax are helpful:

* https://www.w3.org/TR/2012/WD-html-markup-20120329/syntax.html#syntax-attribute-value
* https://www.w3.org/TR/xml/#NT-AttValue

application/backend/src/app.ts
DoxasticFox: Yeah, I tried to make it work by copying the logo into the `application/frontend/public` directory, but fastify-static didn't pick up the new file while Elsa was running. So I had to start Elsa to let it copy the file into `public/`, then stop Elsa, then start it once more. Then I could access the logo in my browser... Which obviously isn't a proper solution.

I don't understand why fastify-static did that. I tried removing `serve: false` from [here](https://github.com/umccr/elsa-data/blob/1ad45f098014cc78dc721aab4fbb97ea84e97e6f/application/backend/src/app.ts#L120) (and also adding `prefix: 'public/'`, because `serve: false` prevented Fastify from establishing a wildcard route which conflicts with [this](https://github.com/umccr/elsa-data/blob/1ad45f098014cc78dc721aab4fbb97ea84e97e6f/application/backend/src/app.ts#L242)). Couldn't get it to work though. It's like it's Fastify builds a directory tree when the application first starts then refuses to update it. Except it's more aggressive about its "caching" than that, because I also tried placing an empty file where the logo would go, then copying the logo over that when Elsa starts. My thinking was that having a placeholder file would help fastify-static set up the route to direct to the file. And it did, I guess. But Elsa still served the empty file after copying the proper logo over it.

mmalenic (approved):

application/backend/src/app.ts
andrewpatto: fair enough - that does sound like the static plugin is weird.

Possibly thinking some more now - if not using built in static server - you might want to set some cache control headers so that the image is not being continuosly fetched (all the other css and js files are aggressively cached). For the image, possibly less aggressive - but maybe an hour?

PR #347, opened 2023-05-21
AG Loader Regex And Refactor

### LINKED ISSUE #328: AG import should extract sample ids from filenames if given regex
The file paths for objects often have a specimen/sample id in them - according to the rules of the lab that did the sequencing. We should allow the dataset config to define a regex (i.e. a capture group) that if matches defines the specimen id.

williamputraintan:
- Regex specimenId (and caseId) from filename when the regex is given in config
- ~Remove exclusiveness constrain on DatasetPatient -> DatasetSpecimen. I think this make sense, that a specimen could belong to a multiple patient (e.g. trios)?~ Refactor on the artifact insertion

### DIFF
--- a/application/backend/config/datasets.json5
+++ b/application/backend/config/datasets.json5
@@ -6,7 +6,8 @@
       description: "A cohort of singleton germline read and variant data",
       loader: "australian-genomics-directories-demo",
       demonstrationStoragePrefix: "s3://agha-gdr-demo-store/10G/",
-      demonstrationSpecimenIdentifierRegex: "^.*(HGddddd).*$",
+      demonstrationSpecimenIdentifierRegex: "^.*(HG\\d\\d\\d\\d\\d).*$",
+      demonstrationCaseIdentifierRegex: "^.*(FAM\\d+).*$",
     },
     {
       uri: "urn:fdc:umccr.org:2022:dataset/10f",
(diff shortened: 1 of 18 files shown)


andrewpatto:
In my head - it would be the Artifact/File that belonged to multiple Specimens - not the Specimen that belonged to multiple patients.. (to support trios). But I haven't completely thought it through so willing to hear arguments

williamputraintan:
> In my head - it would be the Artifact/File that belonged to multiple Specimens - not the Specimen that belonged to multiple patients.. (to support trios). But I haven't completely thought it through so willing to hear arguments

Well, I am just thinking that in case of a `joint.vcf`, individual patients have its own `DatasetPatient` and these `joint.vcf` will belong to a single `datasetSpecimen` (specially taking the specimenId from the filename itself) which will links back to multiple patients.

I think I see what you mean where the artifact that belongs to multiple specimen, but in that case the specimen is bound to the patients not the file specimen itself? 

~ This is just my thoughts, might be wrong

andrewpatto:
So I think the model in my head is - the file/artifact represents the thing that was produced by a run - so is owned by the run/submission batch. It exists independent of whether it is being shared or not and has from 1..* samples columns in it.

The dataset specimen is a record for the sharing of a specific sample out of that file. The dataset specimen would have some of "file sample id" in order to say what column in the VCF the specimen refers to.

The dataset specimen belongs only to one patient

andrewpatto:
The naming of joint called VCFs is problematic in practice.. they really should be named. HG002-HG003-HG004.vcf etc. In reality, the list of sample headers *inside* the VCF content is the canonical "true" list of samples

williamputraintan:
> HG002-HG003-HG004.vcf etc.

Ok will turn the exclusiveness back on.  So I suppose this single VCF will belong to multiple DatasetSpecimen which are:
- SpecimenId - HG002
- SpecimenId - HG003
- SpecimenId - HG004

andrewpatto:
Yes - there'd be one Artifact/File (representing the whole VCF)- and then 3 DatasetSpecimens representing a single sample each from the content.

I mean there's no guarantees the files _will be named this way_ - but I think that's the right model internally to Elsa

andrewpatto (approved):

PR #351, opened 2023-05-29
refactor(frontend): tidy audit event table

### LINKED ISSUE #344: UI issues on audit page
Screenshot 2023-05-17 at 12 58 32 pm

The dropdown checkbox filters is a bit different from anything else - and I don't think it transitions well when we start to add other filters ("Outcome filter") as those don't really fit into a dropdown.

Suggesting we get rid of the dropdown entirely and just have a space above the table with regular daisy controls (which will currently just be the checkboxes)

Also, for some reason - I thought I put an overall query load cursor on the table - so that when you hit the page initially you'd see a loader cursor. But it doesn't seem to be appearing so I must have done something wrong.

mmalenic:
Closes #344.

### Changes
* Remove the filter menu for the audit event table.
* Add check boxes for filtering in between the table heading and table.
* Convert the "all audit events" checkbox to a toggle with a divider to distinguish it from the rest of the check boxes.

### UI Change
image


andrewpatto:
The "all audit events" is not just a toggle - it changes its state depending on the others?
i.e. it will toggle its own state depending on whether some/all of the checkboxes are checked?

application/frontend/src/components/audit-event/filter-elements.tsx
```diff
@@ -0,0 +1,111 @@
...
+              checked={includeEvents.includes("user")}
+              onChange={() => updateIncludeEvents("user")}
+            />
+            <div className="text-xs">User audit events</div>
+          </label>
+        }
+        applyCSS={"font-normal flex content-center items-center grow"}
+        description="Show non-release events"
+      />
+      {allowed.has(ALLOWED_OVERALL_ADMIN_VIEW) && (
```
andrewpatto: I think this is a decision that be expressed in a higher level component and passed down semantically..

so a boolean prop on this component - "showSystemEventOption" (or similar)..

and then that gets passed in the ALLOWED_OVERALL_ADMIN_VIEW..

Its just I wouldn't expect a pretty generic row of checkboxes to have explicit rules about our permissions system..

application/frontend/src/components/audit-event/filter-elements.tsx
```diff
@@ -0,0 +1,111 @@
...
+                onChange={() => updateIncludeEvents("system")}
+              />
+              <div className="text-xs">System audit events</div>
+            </label>
+          }
+          applyCSS={"font-normal flex content-center items-center grow"}
+          description="Show system level events not tied to a user"
+        />
+      )}
+      {allowed.has(ALLOWED_OVERALL_ADMIN_VIEW) && (
```
andrewpatto: Surely "all" is still an option for regular users? Its just the difference between 2 or 3 checkboxes?)

mmalenic:
> The "all audit events" is not just a toggle - it changes its state depending on the others? i.e. it will toggle its own state depending on whether some/all of the checkboxes are checked?

Yes, that's correct, it will affect the state of the other checkboxes. E.g. all release events, vs release events for the current user. Do you think it would be better as a checkbox?

andrewpatto:
I mean more do the checkbox states affect it? If the person is manually just clicking checkboxes - is the "all" toggle being clever about its own state?

application/frontend/src/components/audit-event/filter-elements.tsx
mmalenic: I think difference is that the "all" option allows viewing all audit events, even for other users, which I feel like is an admin-level thing? Should a regular user have access to this? Perhaps the "all" option should be somehow clearer in what it does, or just removed?

mmalenic: Maybe something like: "other users' audit events"

mmalenic:
The checkboxes/toggles don't do anything special, they are just front end on/off switches. The backend decides what to do with the information, i.e. it will return the correct audit events depending on the combination of "filter" information it gets. For example, if it receives "release" and "all", it will return release events for all users (but no other events). I.e. the database decides what to return: https://github.com/umccr/elsa-data/blob/5f517819268109b30fb73f72b2ba7b8545283f54/application/backend/src/business/db/audit-log-queries.ts#L243-L281

Does that make sense? Or are you asking something slightly different?

andrewpatto:
Ok maybe I've misunderstood.. all is not a filter in itself (though I see it in the backend code(?)) - isn't all just the state of all the other checkboxes being ticked?

i.e. the other checkboxes (as a group) have 3 states - none, some, all.. and this bit of the UI is just helping quickly set them/clear them..

application/frontend/src/components/audit-event/filter-elements.tsx
andrewpatto: Ah yes - ok I've misunderstood what "all" means..

andrewpatto: Yes I think possibly ignore the other stuff I've said and just relabel it to "Other Users"

mmalenic: What about something like: "Admin View" with a tool tip that explains what it does? I feel like that way it won't be misunderstood.

andrewpatto: Yeah - so people with the overall admin feature - in the first instance they get a purely personal view of the audit events - identical to other users.. 

but selecting "Admin View" brings in the other rows? Are system events included in things you only see in "Admin View"?

mmalenic: Yes I think including system events probably makes sense. That way there is one less checkbox, and the "admin view" is the only way to view system events (but also allows viewing other users events). I think this lines up conceptually with what an "admin view" is?

andrewpatto (approved):

PR #352, opened 2023-05-31
feat: release viewed audit events

### LINKED ISSUE #90: Add a clever 'View' audit entry when releases are viewed
I would like a 'read' audit entry that allows users to go into and out of 'releases' - but that doesn't generate a huge quantity of meaningless entries.

So looking at some way that 'read' entries are generated - but then automerged somehow (nightly? in a separate ProposedAuditEntry table?) so that if a person went into a release 10 times over an hour - we just get a single audit entry
(Viewed - duration 1 hr).

We will obviously need some time limit such that we can automatically 'finish' off an audit entry - as there is no explicit point at which someone finishes reading a page (unless they log out)

mmalenic:
Closes #90.

### Changes
* Add release viewed audit events when fetching a release.
    * Only allows one view audit event every 30 minutes per release per user.
* Add a general mechanism for timed audit events by caching a key and preventing a new audit events being inserted by using `setTimeout`.
* Move some of the audit event routes to tRPC.


application/backend/src/business/services/audit-event-timed-service.ts
```diff
@@ -0,0 +1,47 @@
+import { injectable } from "tsyringe";
+import { setTimeout } from "timers/promises";
+
+@injectable()
+export class AuditEventTimedService {
+  /**
+   * Audit event which are being awaited on for a given key.
+   */
+  private awaitingAuditEvents = new Set<string>();
```
andrewpatto, line 10: I presume there are implications here then for a load balanced Elsa say? (one with two fargate instances?) - that depending on which service the LB sends it to the behaviour could differ?

(not saying this is a problem - my suspicion is we already have other code like this that would be affected if we tried a multi instance Elsa)

andrewpatto: I guess the question to answer is how badly would this fail if we did run a multi instance?

mmalenic: By multi-instance, do you mean multiple Elsa applications but one database?

application/backend/src/business/services/release-base-service.ts
```diff
@@ -23,13 +25,37 @@ type CodeArrayFields = "diseases" | "countries" | "type";
  * common functionality here and have multiple services.
  */
 export abstract class ReleaseBaseService {
+  /**
+   * 30 minutes.
+   */
+  private readonly VIEW_AUDIT_EVENT_TIME = 1800000;
```
andrewpatto: Couldn't we do this as 30 x 60 x 1000 etc (just makes it easier to alter without others needing to do maths..)

andrewpatto (approved):

application/backend/src/business/services/release-base-service.ts
mmalenic: Yes, good call.

application/backend/src/business/services/audit-event-timed-service.ts
mmalenic: In this case, yes, it would not function correctly. There would potentially be duplicate entries for a user depending on the number of instances. I can add a check to the database to fetch the latest audit entries when inserting and see if there is already one  present. If this sits behind the in-memory cache I don't think it would be too inefficient.

mmalenic:
### Update
* Add database check for existing release audit events when adding a viewed release audit event.
* Fix `Login` navigation tab not showing.

application/backend/src/business/services/audit-event-timed-service.ts
andrewpatto: Yes - I think the db check per instance is a good solution that you've done

PR #353, opened 2023-06-06
Feature/sharer config and refactoring

andrewpatto:
Added a sharer configuration setting that defines which shares the administrator "intends" to enable


andrewpatto:
So this mainly introduces the concept of configured "sharers" - though along the way there were a few other concepts that needed to be introduced.

- Started on the ability to drop in mock AWS services for e2e testing (needed because object sigining will automatically disable itself if it can't discover the object signing service)
- Moving towards the "settings" object being an almost direct pass through of the "configuration" object (will mean less repeating ourselves in declaring these settings)

andrewpatto:
Added some common code for DI the aws clients AND singletons.. just so that we can call the common code from both Elsa entrypoint and test entry..

BUT also this might be where we start to introduce some more clever AWS mocking for E2E (I have only done a minimum here)

andrewpatto:
On the frontend - the sharers that get displayed are now more dynamic and respect the "config" as the master definition of what sharers are *intended* to be allowed. But gives the ability to these to be "broken" and yet still enabled (given these are more dynamic microservices that can be installed independent of Elsa main)

williamputraintan (approved):
Just a minor comment, I think we should expand the is allowed to edit CopyOut bucket? right now people who has the `ViewAsAdmin` role is able to change the bucket name.

DoxasticFox (approved):

mmalenic (approved):

andrewpatto:
> Just a minor comment, I think we should expand the is allowed to edit CopyOut bucket? right now people who has the `ViewAsAdmin` role is able to change the bucket name.

Good question - I was envisaging it as a joint edit between the researchers and admins - but might restrict it to admins only. But yes - certainly not mean for viewAsAdmin people. We should do a separate walk through of all the permissions on all the exposed apis..

PR #355, opened 2023-06-06
Remove MaxMindDb downloader within Elsa

### LINKED ISSUE #333: Move Maxmind DB stuff out of Elsa itself - but allow it to be down as part of Deploy docker builds
Elsa itself can never really deploy an image that has the Maxmind IP database in it - currently it is setup so it can do a download on startup - but I'm not super happy with that.

I think the correct place is in deployment - where someone can take the Elsa base image and easily add in whatever Maxmind db they have (either public or licensed).

So we just need to show how that might be done in the AG demo deploy - with probably the current config settings in Elsa good enough to handle setting the path to where it appears in the docker image.

williamputraintan:
Fix #333


README.md
```diff
@@ -187,6 +187,33 @@ A logo and brand name can be added to Elsa to indicate what organisation is runn
...
+   2. Download the `.mmdb` for the Geo City version (e.g. when using the lite version it is `GeoLite2 City`).
+2. License Keys
+   1. Setup and retrieve the License Keys from the `MY_ACCOUNT` (:bust_in_silhouette:).
+   2. Substitute the key into the following URL `https://download.maxmind.com/app/geoip_download?edition_id=GeoLite2-City&license_key={LICENSE_KEYS}&suffix=tar.gz` and paste the URL to your browser address box (or you can do it via `curl`) to download the `GZIP` file.
+
+The downloaded file will be in GZIP file, you could verify the downloaded file with the SHA256 provided. This database will be updated regularly to provide the most accurate location, at the time of writing this, MaxMind will update this database twice a week.
+
+Extract this data (to obtain the `.mmdb` file) and include this file as part of the deployment.
+
+The maxMindDb is configured using the configuration subsystem described [here](./docs/deployment/configuration.md). An example JSON5 snippet to add a logo and brand name is as follows:
```
andrewpatto: end of sentence is wrong.

application/backend/src/business/services/aws/aws-cloudtrail-lake-service.ts
```diff
@@ -364,14 +364,15 @@ export class AwsCloudTrailLakeService {
...
-      this.logger.warn(
-        "No maxmind db is configured and therefore will not perform IP city lookup."
-      );
+    const maxMindDbPath = this.settings.ipLookup?.maxMindDbPath;
+    if (maxMindDbPath) {
+      try {
+        this.maxmindLookup = await maxmind.open<CityResponse>(maxMindDbPath);
+      } catch (error) {
+        this.logger.warn(
+          "No maxmind db is configured and therefore will not perform IP city lookup."
```
andrewpatto: This is not quite true - "configured maxmind path of ${path} does not contain a readable MaxMind database"

(or something like that)..

application/backend/src/config/config-constants.ts
```diff
@@ -244,6 +241,17 @@ export const configZodDefinition = z.object({
       allowTestRoutes: z.boolean().describe("If test routes should be added"),
     })
   ),
+  ipLookup: z.optional(
+    z.object({
+      maxMindDbPath: z.optional(
```
andrewpatto: if the iplookup is optional - then the path should be mandatory.. (at the moment.. would change if we introduce other databases I guess)

andrewpatto (approved):

PR #357, opened 2023-06-08
Login IP Location Recorded

### LINKED ISSUE #334: Make login route that adds IP address also add Maxmind Location if present
So the login route now adds in details to the audit event saying the source IP address.

Would be great example of the Maxmind IP stuff in action if it also add a field like

'ipLocation' that was the result of a maxmind lookup (happy for it to be very broad - i.e country, state level).

Just needs to prove the point.

williamputraintan:
Fix #334 

Currently embedding into the details prop

Screenshot 2023-06-08 at 18 22 19

Also adding the Flag Icon to the egress record UI.
Screenshot 2023-06-08 at 18 22 40

### DIFF
--- a/application/backend/config/base.json5
+++ b/application/backend/config/base.json5
@@ -2,4 +2,5 @@
   // an ontoserver with MONDO terminology loaded
   ontoFhirUrl: "https://onto.prod.umccr.org/fhir",
   releaseKeyPrefix: "R",
+  ipLookup: { maxMindDbPath: "config/GeoLite2-City.mmdb" },
 }
--- a/application/backend/src/business/services/user-service.ts
+++ b/application/backend/src/business/services/user-service.ts
@@ -20,6 +20,7 @@ import {
 } from "../db/audit-log-queries";
 import { NotAuthorisedModifyUserManagement } from "../exceptions/user";
 import { userGetAllByUser } from "../../../dbschema/queries";
+import { IPLookupService, LocationType } from "./ip-lookup-service";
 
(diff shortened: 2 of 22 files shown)


application/backend/src/business/services/user-service.ts
```diff
@@ -30,11 +31,18 @@ export type ChangeablePermission = {
   isAllowedOverallAdministratorView: boolean;
 };
 
+export type LoginDetailType = {
+  hostname: string;
```
andrewpatto: hostname can be dropped actually.. I thought it was a pass through of some hostname by the load balancer - but in fact it is actually just the name of the server. So is useless in an audit log

andrewpatto (approved):

application/backend/config/base.json5
```diff
@@ -2,4 +2,5 @@
   // an ontoserver with MONDO terminology loaded
   ontoFhirUrl: "https://onto.prod.umccr.org/fhir",
   releaseKeyPrefix: "R",
+  ipLookup: { maxMindDbPath: "config/GeoLite2-City.mmdb" },
```
andrewpatto, line 5: For dev - we just need to source this file and copy it in?
Is it git ignored so we don't accidentailly check it in?

williamputraintan: Added in to `.gitignore`

https://github.com/umccr/elsa-data/blob/1565f5dd3bf1456da858c11ef4f4b05718a94895/.gitignore#L113

PR #360, opened 2023-06-14
Release Participant Management

### LINKED ISSUE #340: Release user management needs permission rules implemented
The user management page in releases allows arbitrary setting of permissions - and has very loose error checking
(i.e. it doesn't even check the things entered are vaguelly an email address - I know this is a hard problem - but we don't even make sure there is an @ symbol)

Tidy this whole page up - which will mean introduction of permission restrictions at the service/backend where this currently none.

Screenshot 2023-05-17 at 9 51 47 am

williamputraintan:
- Convert API -> TRPC
- Guard Participant Role changes
	- Cannot change its own role
	- Cannot add/modify/remove user higher than its own role
- Implement edit and remove button from UI
- Paginate participant table

Screenshot 2023-06-15 at 01 27 10
Screenshot 2023-06-15 at 01 27 37
Screenshot 2023-06-15 at 01 27 59

Fix #340


README.md
```diff
@@ -48,16 +48,18 @@ Permissions bound on each role are as follows.
```
williamputraintan, line 48: Screenshot 2023-06-15 at 09 29 54

andrewpatto:
I think maybe removing the ability to modify via "adding again".. I had only put that in because of a lack of other edit buttons etc. Its not a behaviour that is standard - so I don't think I'd be upset if we got rid of it.

So basically add user will only work for new users - trying to add an existing user will fail with an error message..

andrewpatto:
this is excellent - almost the final piece of the puzzle before we can hand it over to AG folk

application/backend/dbschema/queries/release-participation/releaseParticipantGetAll.edgeql
```diff
@@ -26,16 +27,28 @@ with
...
+        displayName,
+        role := assert_single([is permission::User].releaseParticipant@role) ??
+                assert_single([is permission::PotentialUser].futureReleaseParticipant@role),
+        subjectId := [is permission::User].subjectId,
+        lastLogin := [is permission::User].lastLoginDateTime
+      }
+      order by
+        .role
+      offset
+        <int16>$offset
```
andrewpatto: for consistency - can we check these are the same as all the other paging queries (I think others are int64?)

williamputraintan: Agree, changing it .... some is <int16> (mostly my parts) otherwise it was int64

application/common/elsa-types/schemas-releases.ts
```diff
@@ -217,21 +234,22 @@ export type ReleaseManualType = Static<typeof ReleaseManualSchema>;
...
-  // the internal identifier for this user
+  // the internal identifier (UUID) for this user
   id: Type.String(),
-  // the email address of this user
   email: Type.String(),
   // the role of this user in this release
-  role: Type.String(),
-  displayName: Type.String(),
-  subjectId: Type.Optional(Type.String()),
+  role: Nullable(StringUnion(ReleaseParticipantRole)),
```
andrewpatto, line 241: So is this a mix of Typebox and Zod or something? For trpc can't we pivot to using entirely zod schemas?

andrewpatto (approved):

andrewpatto:
Happy to merge this in - and we can remove the "update via adding" feature later.. unless it is a quick fix

williamputraintan:
>  Happy to merge this in - and we can remove the "update via adding" feature later.. unless it is a quick fix

Doing it in this PR, will update soon!

application/common/elsa-types/schemas-releases.ts
williamputraintan: Yes it's a mix, I think should be possible to move to zod entirely.

williamputraintan:
Won't allow add the same participant twice:
Screenshot 2023-06-15 at 12 13 27

andrewpatto:
"already has" (for the error message) - not just "have"

williamputraintan:
> "already has" (for the error message) - not just "have"
Updating, thanks!

PR #363, opened 2023-06-19
Collection of fixes for AG demo

andrewpatto:
Added mocks for steps/copy out
Make the Elsa client side match the deployment microservices for Copy Out and Object Signing


application/backend/datasets/Smartie/01/manifest.txt
```diff
@@ -1,42 +1,42 @@
...
+2bca1a09e0eccc0a290003b46326ade4	EEE_L001_R1.fq.gz	SUBJ0004
+be7fa630d13b660d9577ef9fa8162fcb	EEE_L001_R2.fq.gz	SUBJ0004
 111ce849044343e2009a3d9c2119534c	EEE.bam	SUBJ0004
 2f5bf0c47c2a8f4599124135a7d2e39b	EEE.bam.bai	SUBJ0004
 11ddcf20ea15911568b7ac28804218a3	EEE.vcf.gz	SUBJ0004
 457799ca7bde302bc11ed53a0694c489	EEE.vcf.gz.tbi	SUBJ0004
 5c5847e6d7307d0a7581f41fe192111f	SUBJ0005.phenopacket.json	SUBJ0005
-2d3ffd5036c6f779eed6e507716b973b	FFF_WGS_L001_R1.fq.gz	SUBJ0005
-991872552a5e899e2f2b2cac3cca1327	FFF_WGS_L001_R2.fq.gz	SUBJ0005
+2d3ffd5036c6f779eed6e507716b973b	FFF_L001_R1.fq.gz	SUBJ0005
```
andrewpatto, line 38: I had the rename the files because the WGS was matching the same regex as my sample ids AAA, BBB

application/frontend/src/providers/logged-in-user-config-relay-provider.tsx
```diff
@@ -34,6 +34,7 @@ export const LoggedInUserConfigRelayProvider: React.FC<{
   const qSettings = {
     refetchOnWindowFocus: false,
     refetchOnReconnect: false,
+    refetchOnMount: false,
```
DoxasticFox, line 37: What does this do?

DoxasticFox: I mean, aside from what it says on the tin? i.e. why is it necessary for Elsa?

DoxasticFox (approved):

application/frontend/src/providers/logged-in-user-config-relay-provider.tsx
andrewpatto: The fetch here of user settings is only meant to run on login - but I'm seeing stuff in the backend log where it seems to be happening more often.. so I was thinking in might be react-query being overly aggressive on refetching..

so here I'm trying to disable *all* refetching

mmalenic (approved):

application/frontend/src/providers/logged-in-user-config-relay-provider.tsx
DoxasticFox: Did it fix it? If not, I remember [William saying React's strict mode causes components to be rendered twice](https://github.com/umccr/elsa-data/issues/272#issuecomment-1491112102).

andrewpatto: Probably not - this was more hit and hope! I did mean all fetches to be disabled - but I don't think this fixes the problem I saw.

I mainly see this during the "Copy Out" job UI - which sets up a background task to do refreshes.. and I think I need to investigate whether that background react axios call is in someway special and might trigger this..

PR #376, opened 2023-06-23
feat: htsget region restriction demo

### LINKED ISSUE #371: Add a simple "genomic region" restriction list in the data custodian htsget sharer
Restrict to
[ none ] (default) 
[ Achromatopsia ]
[ Autism ]
[ Congenital Heart Defect ]

mmalenic:
Closes #371

### Changes
* Adds a demo of htsget region restrictions.

### UI
image


andrewpatto (approved):

application/backend/dbschema/queries/release/applyHtsgetRestriction.edgeql
```diff
@@ -0,0 +1,29 @@
...
+    select exists(r.userRole)
+    or
+    isAllowedOverallAdministratorView
+  )
+
+update r
+filter isAllowed
+set {
+    htsgetRestrictions += <str>$restriction
+}
```
andrewpatto, line 29: not for this PR - but I wonder how we can make this into a boilerplate pattern for all MULTI PROPERTY string fields (for instance). We might have to resort to string replace (horribly)

mmalenic: Yeah, it would be good if there was a flexible way to store string-style properties. I'm not too sure what the best way to store this information is. This could also store the restrictions directly (e.g. as a type that contains chromosome, start and end properties) - which might be the most flexible?

PR #378, opened 2023-06-26
API Enhancement && Fix UI

### LINKED ISSUE #367: William Permissions Bug Hunt June 2023
- [x] ~Admin and~ SuperAdmin should not have access to edit the copy-out bucket name (this should include to guard the API)
- [x] AccessData box should reflect to the config and Data Sharing Control
- [x] When box is disabled (which means `sepia` theme is applied), some text-box/input is still editable
- [x] API for release modification is not guarded with userRole checks (and perhaps also enforce the audit entry when modifying release)

williamputraintan:
- Guard API to only the authorised user ( as well as their respective UI)
- More audit entry on changes on to the 'release'
- Update DaisyUI to V3 and some UI tweaks

Screenshot 2023-06-26 at 22 39 17


application/frontend/src/pages/releases/detail/access-box/manifest-form.tsx
```diff
@@ -1,51 +1,75 @@
...
+              value={
+                "Download TSV" +
+                (!releaseData.activation ? " (Release Must Be Active)" : "")
+              }
+              onClick={() => {
+                setPrepareDownload(true);
+                setTimeout(() => {
+                  setPrepareDownload(false);
+                }, 3000);
+              }}
```
williamputraintan, line 66: Can't find obvious way to show isLoading when using `<form/>` so timer seems to be working fine? (I think google drive also use the same approach ).

<video src="https://github.com/umccr/elsa-data/assets/61998484/08a13aa1-a4c9-4d36-a00b-34e899de2429"/>

andrewpatto: Oh yeah - that looks good. Nice!

andrewpatto:
So the guard you are mainly talking about here is from "AdminView" people? (because the others won't see these panels at all?)

williamputraintan:
> So the guard you are mainly talking about here is from "AdminView" people? (because the others won't see these panels at all?)

Yes, also from the backend as previously some of the API is not guarded with `userRole !== "Admin"` check when doing the `PATCH` request. 

Also the `(Super)AdminView` role will have the same view as the `(Release)Administrator`, but they won't be able to edit on the boxes there.

andrewpatto:
Can we make it so that when "view admins" look at various things - that it fills in the role string "AdminView" (or whatever) into the spot for @role? I think at the moment if I come in as a view administrator - the release tables just leaves the role description/string blank..

andrewpatto (approved):

williamputraintan:
> Can we make it so that when "view admins" look at various things - that it fills in the role string "AdminView" (or whatever) into the spot for @ROLE? I think at the moment if I come in as a view administrator - the release tables just leaves the role description/string blank..

Is it just for the FE purposes? (not stored in Db?)

andrewpatto:
Yes - purely frontend.. if we've now got a string "AdminView" that has some meaning in the backend - even though it does not occur in the database - can we get that string merged into the role strings when they go out to the frontend

PR #381, opened 2023-06-27
fix: hide in progress audit events from the audit event tables

### LINKED ISSUE #364: Long duration audit entries are visible even before completion
Easy to have slipped through as we don't have many long operations - but when doing something like Copy Out that takes many minutes - the partially completed audit entry appears in the UI.

I presume we can just detect "partial" state in the service/query and remove - and we don't need to add an "inProgress" boolean to the db structure.

mmalenic:
Closes #364.

### Changes
* Hides in progress audit events by introducing an `inProgress` property to the `AuditEvent` schema type.

### DIFF
--- a/application/backend/dbschema/audit.esdl
+++ b/application/backend/dbschema/audit.esdl
@@ -53,9 +53,7 @@ module audit {
         property details -> json;
 
         # Whether its expected that this audit event will be updated.
-        required inProgress: bool {
-            default := false;
-        }
+        required inProgress: bool;
     }
 
     # An audit event which can be caused by a user.
--- a/application/backend/dbschema/audit.esdl
+++ b/application/backend/dbschema/audit.esdl
@@ -51,6 +51,11 @@ module audit {
         # bespoke JSON with details of the event
 
         property details -> json;
+
+        # Whether its expected that this audit event will be updated.
+        required inProgress: bool {
+            default := false;
+        }
     }
 
     # An audit event which can be caused by a user.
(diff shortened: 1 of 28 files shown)


DoxasticFox (approved):

application/backend/dbschema/audit.esdl
```diff
@@ -51,6 +51,11 @@ module audit {
         # bespoke JSON with details of the event
 
         property details -> json;
+
+        # Whether its expected that this audit event will be updated.
+        required inProgress: bool {
+            default := false;
```
andrewpatto: I would think this might default to 'true' surely? Or no default at all.

andrewpatto: If I'm laying down an audit event that I expect to mark off later - then by default I am naturally in progress? Until the final save where this then gets set back to false.

mmalenic: I went with false because most of our audit events are instant ones that are never in progress. But that logic also makes sense, so I'm happy to flip these around, or just remove the default?

andrewpatto (approved):

application/backend/dbschema/audit.esdl
andrewpatto: I think remove the default..

andrewpatto:
I also have migrations in progress - so we will clash - but best you merge this one in and then I will regenerate my migrations

williamputraintan (approved):

williamputraintan (approved):

PR #385, opened 2023-06-28
Feature/small geolite

andrewpatto:


andrewpatto:
Constructed a small db of IP ranges that has like CSIRO / Telstra and one US provider. So we can find some actual IP addresses for tests that will work. File is tiny enough to keep in dev and is legally unencumbered.

application/backend/src/business/services/audit-event-service.ts
```diff
@@ -475,7 +477,7 @@ export class AuditEventService {
       orderAscending
     ).run(executor);
 
-    console.log(
```
andrewpatto, line 478: this was just some console log that should be real logs

williamputraintan (approved):

mmalenic (approved):

PR #386, opened 2023-06-28
Feature/dev login to main login page

andrewpatto:
If we want the dev experience to be good - no need to hide these on a magic page.

It is a false security to hide these - they just should not appear on any deployed instance - or if they do appear on a deployed instance the instance should have no real data/features enabled.


andrewpatto:
Screenshot 2023-06-28 at 5 40 56 pm

williamputraintan (approved):
Nice! Like the distinct button colour for each user.

mmalenic (approved):
Nice change - don't have to type `/dev-bm3ey56` anymore.

PR #387, opened 2023-06-28
Fix/empty release activation

### LINKED ISSUE #372: An error message if someone tries to Activate a release but nothing would be released
I think as a business rules we want to prevent people from activating a release - where in practical reality nothing is then released.

So the obvious case is where literally nothing is selected in the Cases box. But also, if they specify no data location OR a data location that doesn't match the dataset.. similarly they might release "reads" only but the dataset has only VCFs.

Anyhow - I think it is probably difficult to make the button gray out on this logic at the moment - but we can certainly detect this when they hit "activate" and we go and make the release manifest. If the release manifest ends up essentially empty of objects - then abort with some sort of expcetion and make sure the exception makes it back to the UI.

mmalenic:
Closes #372

I'm happy for this to be merged after the other PRs.

### Changes
* Shows an error if a release is activated but nothing can be released, if:
    * No cases are selected.
    * No data sources are enabled.
    * There are no specimens.
    * Read or variant data is enabled, but there are no artifacts present for that data type.

### UI
image

### DIFF
--- a/application/backend/src/api/errors/_error.handler.ts
+++ b/application/backend/src/api/errors/_error.handler.ts
@@ -80,14 +80,14 @@ export function ErrorHandler(
 
   // we can't allow a non-error status to come through this code-path - and if we get one - we have to report this
   // as an internal error
-  if (problemResponse.status < 400) {
+  if (problemResponse.status !== undefined && problemResponse.status < 400) {
     problemResponse.type = "about:blank";
     problemResponse.title = "Internal Server Error";
     problemResponse.status = 500;
   }
 
   reply
-    .code(problemResponse.status)
+    .code(problemResponse.status ?? 404)
     .type("application/problem+json")
     .send(problemResponse);
 }
--- a/application/backend/src/business/services/manifests/manifest-master-helper.ts
+++ b/application/backend/src/business/services/manifests/manifest-master-helper.ts
@@ -131,36 +131,71 @@ export async function transformDbManifestToMasterManifest(
         allowedArtifacts.push(a);
     }
 
-    // If read or variant data is allowed, then throw an error if every read or variant artifact is null. That is, there
-    // should be at least some read or variant data artifacts if that type of data is allowed.
-    if (
-      manifest.releaseIsAllowedReadData &&
-      allowedArtifacts.every(
-        (a) =>
-          !a.bclFile &&
-          !a.forwardFile &&
-          !a.reverseFile &&
-          !a.bamFile &&
-          !a.baiFile &&
-          !a.cramFile &&
-          !a.craiFile
-      )
-    ) {
+    specimen.artifacts = allowedArtifacts;
+  }
+
+  checkArtifacts(manifest);
+
+  return manifest;
+}
+
+/**
+ * Checks if artifacts exist across read or variant for the whole manifest if those data
+ * types are enabled. Throws an error if this is not the case.
+ * @param manifest
+ */
+export function checkArtifacts(manifest: ManifestMasterType) {
+  const artifacts = manifest.specimenList.flatMap(
+    (specimen) => specimen.artifacts
+  );
+
+  let checkReadData = (artifact: any) => {
+    return (
+      !artifact.bclFile &&
+      !artifact.forwardFile &&
+      !artifact.reverseFile &&
+      !artifact.bamFile &&
+      !artifact.baiFile &&
+      !artifact.cramFile &&
+      !artifact.craiFile
+    );
+  };
+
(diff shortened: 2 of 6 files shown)


application/backend/src/business/services/manifests/manifest-master-helper.ts
```diff
@@ -32,6 +33,25 @@ export async function transformDbManifestToMasterManifest(
     return false;
   };
 
+  // Check to see if cases/specimens exist to begin with before activating a release.
+  if (manifest.caseTree.length === 0) {
+    throw new ReleaseActivatedNothingError("No cases selected");
```
mmalenic, line 38: One consideration to this is that this will create a lot of failed audit events because it is used in the transactional audit event function. I'm not sure if this matters because these audit events won't be shown anyway?

application/backend/src/api/errors/_error.handler.ts
```diff
@@ -80,14 +80,14 @@ export function ErrorHandler(
 
   // we can't allow a non-error status to come through this code-path - and if we get one - we have to report this
   // as an internal error
-  if (problemResponse.status < 400) {
+  if (problemResponse.status !== undefined && problemResponse.status < 400) {
```
andrewpatto, line 83: how are we getting undefined statuses?

andrewpatto: (as in what has changed - or has this always been broken for certain error classes?)

andrewpatto (approved):

application/backend/src/business/services/manifests/manifest-master-helper.ts
```diff
@@ -111,6 +131,34 @@ export async function transformDbManifestToMasterManifest(
...
+      );
+    }
+    if (
+      manifest.releaseIsAllowedVariantData &&
+      allowedArtifacts.every((a) => !a.vcfFile && !a.tbiFile)
+    ) {
+      throw new ReleaseActivatedNothingError(
+        "Variant data is enabled but there are no variant data artifacts"
+      );
+    }
```
mmalenic: I'm quite unsure about this logic. Should there be at least some artifacts for a given specimen if that data type is enabled (i.e. reads or variants)? This could also be a bit less strict, and be applied outside of the loop - i.e. there should be at least some reads/variants data across the whole manifest, and not for each specimen?

andrewpatto: Yes for the moment I think I'd prefer that - the looser version. I am happy if only half the samples have reads - and therefore only half the samples get released (with no other error)
(for the moment)

application/backend/src/api/errors/_error.handler.ts
mmalenic: I changed the status to be optional in this PR - technically this is allowed in the 7807 error spec. I did this so that it was a bit more straight-forward to render detailed error messages without a status code, although I can find another way to do this if you think the status should be always required?

application/backend/src/business/services/manifests/manifest-master-helper.ts
andrewpatto: I don't see a problem with attempting this and having a failed audit event.. if it becomes a problem we can work on tidying up the UI (maybe some more pre-emptive warnings)

andrewpatto:
Does this also handle the situation of files that are only in S3 - but the user picks GS (and not S3)?

application/backend/src/business/services/manifests/manifest-master-helper.ts
mmalenic: Okay, I'll change this to be less strict.

application/backend/src/api/errors/_error.handler.ts
andrewpatto: Nah that sounds fine. I just didn't know if something major had changed to prompt this

andrewpatto:
> Does this also handle the situation of files that are only in S3 - but the user picks GS (and not S3)?

So this might be the very general most of exceptions right at the end - if after doing all the other processing - the literal list of URLs in the manifest is empty - then that should be a fail. (not sure if I can see that logic in here)

mmalenic:
> > Does this also handle the situation of files that are only in S3 - but the user picks GS (and not S3)?
> 
> No, it only detects if no data sources are selected.

Yes, I forgot to add that case, I'll add that in too.

mmalenic:
### Update
* Move check for reads and variants data outside of loop.
* Add check for empty artifacts overall.

andrewpatto (approved):

DoxasticFox (approved):

PR #388, opened 2023-06-29
Add a user interface to view previously run jobs for a release and their details

### LINKED ISSUE #365: Add a user interface to view previously run jobs for a release and their details
We record the status/details of all the jobs that run in the db - but we give the admins no mechanism to view them.

We should add a new sub detail page called Jobs for each release - that contains a view of the jobs and basic details such as whether they suceeded and when they ran. I mean, to some extent this is information in the audit event - but might have more details and is all in one spot.

DoxasticFox:
Closes https://github.com/umccr/elsa-data/issues/365. Mostly copy/pasta from @mmalenic's work on audit events.

Untitled


andrewpatto:
As a general rule - I think the created/started columns will be effectively identical - so I wonder if there is a way we can show only the "created" (say), but with the started shown some other way (tooltip on a "info" icon) only when it differs by more than a minute.

I mean I think jobs can fail by "not starting" - but can't think of the exact circumstances in which they would differ as times..

andrewpatto:
basically I don't know the answer here - but for 99% of the time showing both is not going to be particular useful I think

andrewpatto:
I mean we aren't short for horizontal space in the table so I guess it doesn't matter..

andrewpatto:
We might need a way to get the job type specific fields out into the front end - not just the 'messages'.

For instance - the detail of which Stack was deleted by the uninstall operation comes from the `awsStackId` field - and might be useful if people are looking at the historical record of jobs.

Its possible we don't need to extend the modelling all the way to the front end though - possibly the backend queries just bundle that info into the first lines of the messages? Or some other way it just passes down the field data for "generic" display in the frontend.

andrewpatto:
Otherwise is looking good

DoxasticFox:
Screenshot_20230703_195652

application/backend/src/api/routes/internal-trpc/input-schemas-common.ts
```diff
@@ -12,6 +12,10 @@ export const inputReleaseKeySingle = z.object({
   releaseKey: inputReleaseKey,
 });
 
+export const unorderedInputPaginationParameter = z.object({
+  page: z.number().positive().default(1),
+});
+
```
andrewpatto, line 18: good! I was adding a paged TRPC and realised I also need this..

application/frontend/src/index-router.tsx
```diff
@@ -36,12 +38,16 @@ import {
...
+  features: Set<string>;
+};
+
 /**
  * Create the complete set of routes for the application.
  *
  * @param features the set of features to enable in the UI
  */
-export function createRouter(features: Set<string>) {
+export function IndexRouter({ features }: IndexRouterProps) {
```
andrewpatto, line 50: Nice! Not sure why I did it as a function in the first place..

andrewpatto (approved):

williamputraintan (approved):

application/frontend/src/index-router.tsx
DoxasticFox: Probably because react router's requirement of using `createBrowserRouter(createRoutesFromElements(...))` is awkward and forces the use of functions to generate JSX objects somewhere. The documentation for the return types is also so-so.

I actually fiddled with this file more than I needed to in this PR because I initially attempted to use a hook (`useUiAllowed`) to dynamically show/hide the "Jobs" breadcrumb. That didn't work the way I'd hoped, but I still kept some of the changes.

PR #391, opened 2023-06-29
Update Data Egress Record Refinement 

### LINKED ISSUE #382: Scheduled Task Service (for Data Egress Loader)
Add template for cron job task with the first use case of:

- Data Egress record loader needs to be triggered everyday to have the latest data as possible. This also allow each on-demand egress loader to be faster as it will have a lesser interval time to query.

williamputraintan:
closes #382 

- Add scheduled task to update data egress daily (at 6am)
- Refine UI (sorting table and styling)
- Egress loader is now idempotent and make use of `eventId` from AWS as the id record
- Make startDate of query to be at 00:00 each day, and have the scheduler to run at ~6am~ based on configured time. CloudTrailLake query interval is within a day, so querying with eventTime filter from 6am or 12am in the morning will have the same cost. 

- Make egress timestamp start day -1 hour before the last query this is because, egress event may take up to (or even more) 15 minutes based on [AWS doc](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-lake.html), so it is possible that an egress happen on 5.50am and they aren't there yet when the update was done at 6am (the `lastQueryTimeStamp` will have 6am as record). And if the next query only start from 6am (prev days), it will miss these event records so by moving to start querying from an hour before the last query could prevent the miss.
	
	
Screenshot 2023-06-29 at 17 59 43


application/backend/src/entrypoint-command-start-web-server.ts
```diff
@@ -118,9 +122,15 @@ export async function startJobQueue(config: any) {
           workerData: config,
         },
       },
+      {
+        name: jobEgressUpdateFileName,
+        cron: "0 6 * * *", // Every 6am
```
andrewpatto: can't we get this to literally come from the Elsa config/settings?

andrewpatto: by the time the Bree is started we have resolved all the config/settings

williamputraintan: Ah, yes that would be better!

application/backend/src/entrypoint-command-helper.ts
```diff
@@ -87,3 +91,34 @@ export async function getFromEnv(): Promise<{
     redactedConfig: JSON.parse(JSON.stringify(config)),
   };
 }
+
+/**
```
williamputraintan: So I am thinking to create a `SYSTEM` so there will be the normal user logging if an action  taken by the user (e.g. scheduler to update egress-records).

Thoughts?

mmalenic: I think this makes sense.

mmalenic (approved):

application/backend/src/entrypoint-command-helper.ts
williamputraintan: Well alternatively for audit event purposes, I see we could use the `systemAuditEventPattern` but I guess need to change a bit on the `releaseLastUpdatedReset` (for notifying the recent changes), and `bucketNameMap` as this activities need some userRole/isAllowed checks before able to access the resource.

andrewpatto: I'm slightly hesitant to introduce a System user - it was one option to handle other system events - but we went down the explicit audit event pattern.

andrewpatto: So from an audit event perspective I think it is not needed.

I do understand though that some of our services are structured in a way that requires a real user. I think possibly we can fix that with some refactoring though..

williamputraintan: Yeah, I was a bit hesitant too here. Hmmm, well for the egress scheduler will just need to move the `getBoundaryInfoWithThrowOnFailure` out from the `getInstalledAccessPointResources` function. 

But I am thinking if we could move most of the `userRole` or the `getBoundaryInfoWithThrowOnFailure` checking if possible up to the router level so the service could use other services without the restriction?

DoxasticFox (approved):

williamputraintan:
Update 11/07:
- Add data-egress helper function
- Organise tests and business-services into folders
- Break egress record updater into its own service from aws-cloudtrail-lake
- Remove the fake `SYSTEM` user

andrewpatto (approved):

application/backend/config/base.json5
```diff
@@ -3,4 +3,12 @@
   ontoFhirUrl: "https://onto.prod.umccr.org/fhir",
   releaseKeyPrefix: "R",
   ipLookup: { maxMindDbPath: "config/Geoacumen-Country.mmdb" },
+  feature: {
+    enableConsentDisplay: true,
+    enableCohortConstructor: true,
+    enableDataEgressViewer: {
```
andrewpatto: just a quibble - the intent for the feature list was just a set of boolean flags.. (possibly we'd have some env mechanism to control them as well)..

I'd like to keep the feature flag stuff sepearate from the actual config of the features. Like feature flags are a mechanism for us as developers/deployers to control what is active/viewable/present - whereas the config itself is for the users to set/use.

So can we have a dataEgress config section - with whatever config you want.. and 100% it might be that without that config present the data egress is disabled.

But separately the data egress feature flag is for broad scale disabling the concept from the stack entirely

PR #392, opened 2023-06-30
Get E2E working again

### LINKED ISSUE #390: Get E2E tests to detect react(only?) error console output
We sometimes won't notice React errors that get dumped to the javascript console.

I believe there is a relatively generic technqiue that you can use in Playwright that can parse the console output and we could use it to detect unexpected errors that occur during any e2e test.

Investigate whether this is possible and give a demo of the technique (to be used going forward as we develop more e2e tests)

andrewpatto:
No wait after e2e test

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -64,7 +64,8 @@ jobs:
       # install the EdgeDb CLI
       - uses: edgedb/setup-edgedb@v1.2.2
         with:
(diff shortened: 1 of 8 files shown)


williamputraintan (approved):

mmalenic (approved):

andrewpatto:
Closes #390

PR #394, opened 2023-06-30
Release Password Viewer

### LINKED ISSUE #383: Add a reveal password button/display into the Information Box
We need to allow everyone to be able to see the releasePassword associated with a release.

Currently this is set - and used for the zip files - but not displayable to anyone.

I think it should be retrieved on a separate TRPC call - and we should remove it from the base Release GET.

We also then need somewhere where someone can do something to fetch it and have it displayed for a minute in a copy textbox.

williamputraintan:
Closes #383 

<video src="https://github.com/umccr/elsa-data/assets/61998484/19d8f0db-2019-4757-85eb-4d0f52592e07" />

### DIFF
--- a/application/backend/src/api/routes/internal-trpc/release/release-router.ts
+++ b/application/backend/src/api/routes/internal-trpc/release/release-router.ts
@@ -1,4 +1,5 @@
 import { z } from "zod";
+import { ReleaseActivationPermissionError } from "../../../../business/exceptions/release-activation";
 import {
   calculateOffset,
   internalProcedure,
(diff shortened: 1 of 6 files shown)


application/backend/src/api/routes/internal-trpc/release/release-router.ts
```diff
@@ -41,6 +42,25 @@ export const releaseRouter = router({
 
       return await ctx.releaseService.get(user, releaseKey);
     }),
+  getReleasePassword: internalProcedure
+    .input(inputReleaseKeySingle)
+    .mutation(async ({ input, ctx }) => {
+      const { user } = ctx;
+      const { releaseKey } = input;
+
+      return await ctx.auditEventService.transactionalReadInReleaseAuditPattern(
```
williamputraintan, line 51: Am thinking of moving the audit record insertion to the route level instead in the service level (except for action done by the system)? Well I think it should log who did what, and usually each API request represent that.

If say implementing the auditPattern inside the `.getPassword` function, this function is called on this API and when the system need to know the password for the encryption `tsv`. I don't think when encrypting the password should also have this (getPassword) audit record and the "create-tsv" audit record.

DoxasticFox (approved):

andrewpatto (approved):
Happy to have this merged and we can revisit the audit pattern/service thing later (as this is perfectly functional)

PR #398, opened 2023-07-03
Feature/refactor user queries

### LINKED ISSUE #384: Move GET users route from REST to TRPC
Seems like a simple one- we have a REST Users route that just is basically a paged get.

Can move this over to TRPC

andrewpatto:
Converted the user stuff from REST to TRPC
Changed user queries from typescript to edgeql


application/backend/dbschema/queries/permission/userGetAllByUser.edgeql
```diff
@@ -1,36 +1,36 @@
...
   # we may instead get all the other users
-  canSeeOtherUsers := (
-    select permission::User {
-      isAllowedOverallAdministratorView
-    }
-    filter .id = <uuid>$userDbId
-  )
-  .isAllowedOverallAdministratorView or <bool>$isSuperAdmin,
+  canSeeOtherUsers := <bool>$isSuperAdmin
+     or (
```
andrewpatto, line 13: added the ability for this query to be made with no user - in which case it returns everyone and there is no concept of a caller

application/backend/src/api/routes/trpc-bootstrap.ts
```diff
@@ -80,8 +80,12 @@ const isSessionCookieAuthed = middleware(async ({ next, ctx }) => {
   // Parse pagination from req
   const pageSize = currentPageSize(ctx.req);
 
+  /*
```
andrewpatto, line 83: @williamputraintan I have removed all the permissions from the AuthenticatedUser (the one that is stored in the session cookie) - so the comparison is not longer valid. It is possible we could compare against the front end cookies? Anyhow have disabled for now

application/backend/src/business/services/release-base-service.ts
```diff
@@ -164,6 +142,10 @@ export abstract class ReleaseBaseService {
       userRole: role as UserRoleInRelease,
       isActivated: !!boundaryInfo.activation,
       isRunningJob: !!boundaryInfo.runningJob,
+      isAllowedOverallAdministratorView:
+        boundaryInfo.isAllowedOverallAdministratorView,
+      isAllowedCreateRelease: boundaryInfo.isAllowedCreateRelease,
+      isAllowedRefreshDatasetIndex: boundaryInfo.isAllowedRefreshDatasetIndex,
```
andrewpatto, line 148: Note that the release Boundary check now also returns the user permissions live from the db as an extra field

application/backend/tests/service-tests/ag-s3-import.test.ts
```diff
@@ -75,8 +75,6 @@ describe("AWS s3 client", () => {
 
     ({ existingUser: user, edgeDbClient: edgedbClient } =
       await beforeEachCommon());
-
-    await blankTestData();
```
andrewpatto, line 79: this was deleting the users that were created in beforeEachCommon - so the test users were actually different User records

andrewpatto:
Closes #384

DoxasticFox (approved):

williamputraintan (approved):

application/backend/src/api/routes/trpc-bootstrap.ts
williamputraintan: Noted!

I guess, one way is to always send a cookie that reflect on the db, and let the FE decide if they are out of sync with the BE.

williamputraintan: At the very least, the API will throw an error / return an empty array when they are no longer authorised.

PR #400, opened 2023-07-04
An Initial OWASP evaluation docs

### LINKED ISSUE #196: Web Application Security Checklist
- [ ] Broken Access Control
	Currently the application does not check for all API if user is authorise to access datasets/release that belongs to them. This tick should implement authorisation checks, some API may already had a TODO note.
- [x] Apply CSRF token
	The app is using a cookie based authentication and so it is prone to a CSRF attack. Implementing a CSRF token and/or cookie flag to ensure cookie is sent to the same site

williamputraintan:
I think we could get this merge as handover is coming soon, just recently modify the OWASP evaluation based on what we have currently.

Happy for any comments/feedback here.

Closes #196


SECURITY.md
```diff
@@ -15,3 +15,69 @@ Use this section to tell people how to report a vulnerability.
...
+
+Additionally, we have enhanced logging by implementing data egress logs, as data access is a primary function of the application. If enabled by the software administrator (which currently only works in the AWS deployment), this feature tracks the data egressed by each researcher. For example, when a researcher generates pre-signed URLs for a dataset, accessing those URLs triggers the retrieval of information from the storage server, which is then stored in the application database. Data custodians can monitor and log these activities, and certain events, such as data downloads from foreign IP addresses or multiple downloads, are flagged to highlight potential anomalies and to track egress costs.
+
+### 10. Server-Side Request Forgery (SSRF)
+
```
williamputraintan: Thinking something related to the AWS deployment where we have configured for some of the security group rules that could prevent SSRF. Thoughts? 

Many of the example was related to fetching on external links, but I don't think we do any external link fetching?

Or perhaps something to do with url/file validation for the DAC page?

SECURITY.md
```diff
@@ -15,3 +15,69 @@ Use this section to tell people how to report a vulnerability.
...
+For file encryption, to ensure the security of active presigned URL files, we encrypt them with a randomly generated UUID, which is stored in the database. This prevents participants from storing unencrypted presigned URLs locally. Participants can obtain the password for accessing these files through the application. Currently, the password is stored as a string as the purpose of this feature is solely to protect unencrypted files stored on local devices.
+
+### 3. Injection
+
+In Elsa, we have implemented various measures to minimize the risk of injection attacks, particularly by carefully handling user-supplied input. Most of the inputs in our application are checkboxes, which significantly reduce the chances of injection vulnerabilities. Furthermore, we have eliminated the use of raw SQL queries in our database interactions. Instead, we leverage the [EdgeDb Client Library](https://www.edgedb.com/docs/clients/js/index) and a predefined [EdgeQL](https://www.edgedb.com/docs/edgeql/index) template for our database queries. These client libraries and EdgeQL templates follow strict casting rules, ensuring data integrity before any interaction with the underlying SQL database.
```
andrewpatto: I don't think the Checkboxes sentence is needed - unless we literally were not having *any* text input at all.

I think we just state we are using the EdgeDb client libraries and those prevent injection attacks due to strongly typed parameters to the Edgeql (so pretty much along the lines of what you are saying).

SECURITY.md
andrewpatto: Unfortunately you can only do security egress rules by IP address - so it might be difficult for us to whitelist.

The true best case is if we can get Elsa to run on a private subnet with *no* egress.. that takes out heaps of attack vectors.. it even wrecks some supply chain attacks because they can't contact their home base to open up a control channel. 

Perhaps we should do an investigation - what external access does the core of Elsa actually require (other than AWS calls for instance)

mmalenic: I feel like Elsa isn't really doing much external fetching anyway? I think SSRF is mostly about that, but I guess internal access to resources also matters.

mmalenic (approved):

SECURITY.md
andrewpatto: Yes I can't think of any deliberate external fetching - but there are lots of ones you don't think about (fetching certs to verify JWT etc). And then every AWS call is an external call unless we set up an endpoint.

It would be great if we got a handle on what calls/externals are needed and work out if there are workarounds.

As I said, it would be fantastic if Elsa literally couldn't access external URLs and still worked - as that then is a hard counter to lots of modern threats

andrewpatto:
I know its markdown so it all works as is - but can we wrap the lines at < 100 cols? It disturbs me when the text goes the width of my browser/editor..

SECURITY.md
```diff
@@ -15,3 +15,69 @@ Use this section to tell people how to report a vulnerability.
...
+
+With regards to deployment configuration, we have established a crafted template for safe deployment on AWS. This template is designed to provide a safe and reliable deployment setup, incorporating preconfigured settings that we believe are secure. To access our Elsa CDK deployment template please visit https://github.com/umccr/elsa-data-aws-deploy.
+
+### 6. Vulnerable and Outdated Component
+
+On our repository, we currently have set up a [dependabot](https://github.com/dependabot) to notify for any alarming version of the library used in the project.
```
andrewpatto: Also I think our Docker images are scanned by github on publish (or if not they should be).

If we do progress then "works without external network access" - then I would mention it here as it mitgates vulnerable components via supply chain attacks

SECURITY.md
```diff
@@ -15,3 +15,69 @@ Use this section to tell people how to report a vulnerability.
...
+
+Another important aspect of our design is the support for HTSGET, a data sharing mechanism. HTSGET enables the data custodian to control access to specific alignments within genomic files, ensuring that participants can only access the relevant data instead of providing access to the entire file. For more information on HTSGET, please visit the official documentation at http://samtools.github.io/hts-specs/htsget.html.
+
+### 5. Security Misconfiguration
+
+The application configuration is stored in the dedicated configuration folder, which is copied into the container image during deployment. This configuration folder is organized into sub-topics, such as super-admins and the list of datasets, and is parsed using the a [convict library](https://github.com/mozilla/node-convict/tree/master). By centralizing the high vulnerability configuration in this folder, we minimize the risk of accidental modifications. It is important to note that any changes to the configuration can only be made when rebooting the application. This deliberate approach provides an added layer of protection, reducing the chances of unintended configuration modifications.
```
andrewpatto: we don't use convict at all any more. We use zod schema.

SECURITY.md
williamputraintan: Removing it!

williamputraintan:
> I know its markdown so it all works as is - but can we wrap the lines at < 100 cols? It disturbs me when the text goes the width of my browser/editor..

Sure! sorry I got my editor text-wrap

andrewpatto (approved):
I'm happy with the content - but leave the PR unmerged and I'll do some small language fix ups on the branch directly

DoxasticFox (approved):

PR #404, opened 2023-07-07
Don't use UUIDs as passwords (hard to copy/paste)

andrewpatto:
Also better defaults for AG


andrewpatto:
Nothing wrong with UUIDs in general - but given this is a password we expect them to copy/paste it is nicer if it is just plain old characters.. also they may need to extract the zip on a HPC - in which case they may need to literally type the password in manually on another system..

mmalenic (approved):

PR #413, opened 2023-07-10
Small micro fixes for researcher view

andrewpatto:
Access methods should be enabled based on presence of config AND of whether they are enabled (this is different logic than for Admins)


andrewpatto:
@williamputraintan I've got a hour long meeting a 4pm.. if you have time can you check the service test here that's failed - it seems like it was expecting a sort order in the service test that isn't present. So some random peturbation in the backend has made the test fail. Maybe make the test sort independent?

andrewpatto:
I can't think of anything I've changed in the frontend here that would cause the AG import backend the start failing!

williamputraintan (approved):

mmalenic (approved):

PR #416, opened 2023-07-10
fix(frontend): activation button should be disabled when data is fetching

### LINKED ISSUE #415: Activate and deactivate buttons should possibly be on isFetching not isLoading
I think the distinction between the two is whether data has already been fetched.

But for the purposes of activate and deactivate buttons - we don't care whether there is already existing ReleaseData - what we want to prevent is people clicking these buttons whilst the operations are still happening (including fetching the "new release data").

So I think they should be disabled in the most cautious way - whatever that is.. so possible the isFetching on both the mutation AND the get release data?

(I have seen the buttons become active again before I feel the UI has fully reflected the new state - which means that if I did click it would be trying to activate an active release etc)

mmalenic:
Closes #415

Hopefully this disallow pressing the activation buttons when the queries are in progress.

### Changes
* Make the activation and deactivation buttons disabled when data is fetching, or a mutation is running.

### DIFF
--- a/application/frontend/src/pages/releases/detail/information-box.tsx
+++ b/application/frontend/src/pages/releases/detail/information-box.tsx
@@ -10,7 +10,7 @@ import { EagerErrorBoundary } from "../../../components/errors";
 type Props = {
   releaseKey: string;
   releaseData: ReleaseTypeLocal;
-  isFetching: boolean;
+  releaseDataIsFetching: boolean;
 };
 
 /**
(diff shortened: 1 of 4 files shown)


andrewpatto (approved):

application/frontend/src/pages/releases/detail/information-box.tsx
```diff
@@ -10,6 +10,7 @@ import { EagerErrorBoundary } from "../../../components/errors";
 type Props = {
   releaseKey: string;
   releaseData: ReleaseTypeLocal;
+  isFetching: boolean;
```
andrewpatto: possibly name as releaseDataIsFetching?

andrewpatto: given there are potentially other higher level fetching going on - and certainly some mutations at this level also isFetching

mmalenic: Yes, good idea.

PR #418, opened 2023-07-11
Hide Password in Audit

### LINKED ISSUE #410: Audit Event to get password for release can audit the 'current' password - but only first characters
It is possibly useful to be able to know what password was given out when (if we did password rotation for instance) - but would suggest we perhaps only audit the first few characters to the audit entry

Either 
'ABCD********'..

or

'ABCD...'

williamputraintan:
Closes #410 

- Hide Password in Audit
- Change Mutate to Query (Ref: [Slack](https://umccr.slack.com/archives/C03F7BHNJ3X/p1688961935519719))
- Move auditing back to service-layer

Screenshot 2023-07-11 at 11 53 51


e2e/playwright.config.ts
```diff
@@ -72,7 +72,7 @@ const config: PlaywrightTestConfig = {
       AWS_REGION: "ap-southeast-2",
       ELSA_DATA_VERSION: "e2e",
       ELSA_DATA_BUILT: "now",
-      ELSA_DATA_REVISION: "abcd",
+      ELSA_DATA_REVISION: "ABCDEFGHIJKL",
```
andrewpatto: ??? (this wasn't a password)

application/backend/tests/test-elsa-settings.common.ts
```diff
@@ -109,7 +109,7 @@ export const createTestElsaSettings: () => ElsaSettings = () => ({
   oidc: {
     issuer: ciLogonIssuer,
     clientId: "12345",
-    clientSecret: "abcd" as any, // pragma: allowlist secret
+    clientSecret: "ABCDEFGHIJKL" as any, // pragma: allowlist secret
```
andrewpatto: this also isn't a release password so would leave as is

e2e/playwright.config.ts
williamputraintan: sorry, used auto replace which replaces this eventually.

andrewpatto (approved):

PR #420, opened 2023-07-11
Fix Unauthorised Message

### LINKED ISSUE #407: Credential expiry detect seems to kind of work but with no message
Screenshot 2023-07-10 at 10 48 19 am

This was when I came back after a weekend - having been logged in on Friday.

It seems to have successfully noticed the expired login and blurred out - which looks good and is the desired behaviour.

But the message is empty.

Also though - in another tab in the same firebox - logged in on the same Friday - a different account was not blurred at all (it just fails in the backend queries on the releases table). Also, not the end of the world - but wonder if the detection of credential expiry only works in some pages?

williamputraintan:
Closes #407 

So for the blank message, it is just the layout of TRPC vs REST error was in a different format. So it was blank for the message.
Screenshot 2023-07-11 at 16 26 44

For the browsing here and there after token has expired, after investigating it seems mainly due to csrf token fall into internal server error (500), as it dumps anything else to 500. The default trpc error was 403 error, but I think we could make it 401 and just assume they are no longer authorised. 

A bit more detail: So when the session-cookie expired in the backend, the CSRF-protection  check will also fail. But the order of checking is CSRF token then it goes to our normal auth. Since the last implementation 500 was returned, it was not detected by the FE (as an authentication failure).

### DIFF
--- a/application/backend/src/api/errors/_error.handler.ts
+++ b/application/backend/src/api/errors/_error.handler.ts
@@ -36,7 +36,7 @@ A problem details object can have the following members:
 
 import { FastifyError, FastifyReply, FastifyRequest } from "fastify";
 import { Base7807Error, Base7807Response } from "@umccr/elsa-types/error-types";
-import { NotAuthorisedCredentials } from "./authentication-error";
+import { NOT_AUTHORISED_MESSAGE } from "./authentication-error";
 
 export function ErrorHandler(
   error: Error,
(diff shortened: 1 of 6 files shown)


application/backend/src/api/errors/_error.handler.ts
```diff
@@ -66,6 +67,13 @@ export function ErrorHandler(
         problemResponse.status = 400;
         problemResponse.detail = error.message;
         problemResponse["validation-errors"] = (error as any).validation;
+      } else if (
+        (error as any).code == "FST_CSRF_MISSING_SECRET" ||
+        (error as any).code == "FST_CSRF_INVALID_TOKEN"
+      ) {
+        problemResponse.title = new NotAuthorisedCredentials().message;
```
andrewpatto: if the message is a static string - then should make it static on the class (rather than require a new)

andrewpatto (approved):

PR #421, opened 2023-07-11
Fix some of the wait cursor logic for cases box

### LINKED ISSUE #408: Wait cursors for Case box search aren't correct
The case box search seems to have various phases - and on first startup (i.e when the database/indexes are cold) - takes a bit of time. Which is all fine - it doesn't need to be super high performance. But as per the attached movie - it is not clear to the user that the search is still in progress as this happens. Not sure what sort of wait cursors are appropriate and where (because it definitely seems to go through multiple phases) - but needs sorting out in some way.

https://github.com/umccr/elsa-data/assets/1973827/c744279f-e849-4e1a-944e-998f484602a3

### LINKED ISSUE #412: Add a summary of selected cases to the cases box
ONLY IF ABLE TO BE DONE AT SPEED - but it would be useful if the admins could see a summary of what is involved in the release as currently selected.
 e.g. (totally a guess - NOT A SPEC!)

Screenshot 2023-07-10 at 2 57 59 pm

andrewpatto:
Also convert it all to TRPC


application/backend/src/app-router.ts
```diff
@@ -15,10 +15,10 @@ export const appRouter = router({
   dac: dacRouter,
   auditEventRouter: auditEventRouter,
   datasetRouter: datasetRouter,
+  release: releaseRouter,
```
andrewpatto, line 18: for consistency am dropping the word router off the field name
(will do the same for dataset and auditEvent in another PR)

application/frontend/src/pages/releases/detail/cases-box/cases-box.tsx
```diff
@@ -43,105 +78,95 @@ export const CasesBox: React.FC<Props> = ({
   isEditable,
   showConsent,
 }) => {
+  // a quasi state for just the UI that tracks if we think the entire set is checked or unchecked
+  // (basically we are guessing it is indeterminate unless we get a strong signal via button press
+  // that everything is cleared or set)
   const [isSelectAllIndeterminate, setIsSelectAllIndeterminate] =
     useState<boolean>(true);
 
-  const [isLoading, setIsLoading] = useState<boolean>(true);
```
andrewpatto, line 53: I suspect the text searching isLoading and dataQuery.isLoading as two separate states was what was occasionally making the UI seem delayed/out of sync (though agree it is hard to replicate @DoxasticFox ).

Also, I think the query invalidate() was possibly too harsh - possibly causing extra fetches of the release data at the same time as the cases query.

andrewpatto:
Am not 100% certain this actually completely fixes things - but it is (slightly) better than before..

williamputraintan (approved):

DoxasticFox (approved):

PR #431, opened 2023-07-14
feat: email service update and release activation/deactivation emails

### LINKED ISSUE #414: Email researchers on release activate and deactivate
Whilst we have an email mechanism - we don't actually send any emails!

First cab I think is that everyone involved in a release should be notified when the release is activated and deactivated.

mmalenic:
Closes #414.

### Changes
* Updates the email service so that it can send `pug` email templates.
    * Templates are under the `emails` directory, and can be overwritten in the docker build.
    * Config allows passing a template dictionary which can overwrite or define new template `locals`.
    * Config allows setting the `from` email address and display name.
    * It is possible to retry sending an email when it fails, although I'm not sure if this is useful (currently the system only attempts to send an email once).
    * There is a failed audit event if the email couldn't be sent, otherwise it is a successful audit event.
* Emails all participants of a release when it is activated or deactivated (including potential users).
* Small dev config refactor.
    * Email sending is disabled by default when in dev mode - instead emails are rendered and the HTML can be viewed.
* Add email service unit tests.
    
### Email
* An email in my inbox (could potentially add the branding icon as the profile picture?):
image


application/backend/config/dev-localhost.json5
```diff
@@ -68,8 +68,12 @@
     // we have a permanent tmp bucket with a generous expiry lifespan shared for localhost dev
     tempBucket: "elsa-data-tmp",
   },
-  mailer: {
```
andrewpatto, line 71: ruling out paper I see?

mmalenic: Pretty much - no use for paper here.

application/backend/src/business/services/email-service.ts
```diff
@@ -0,0 +1,195 @@
...
+      from: Address,
+      to: string
+    ) => Promise<any>,
+    from: Address,
+    to: string
+  ): Promise<any> {
+    // Not sure that there's much point trying to send an email more than once, but I could be wrong.
+    const tryCount = 1;
+    let [result, tried] = [undefined, 0];
+    try {
```
williamputraintan: For the try-catch with the audit record, I started to use the `systemAuditEventPattern` which I think simplify the try-catch and success/fail outcome record.

Seems that a success will always be inserted, even if it fails?

mmalenic: Good catch, yes it's not returning/throwing early when an error is caught. I'll look at `systemAuditEventPattern` too.

DoxasticFox (approved):

andrewpatto (approved):

williamputraintan (approved):

PR #432, opened 2023-07-17
Move most cookie storage  to `useContext`

### LINKED ISSUE #401: Reestablish a check to tell the user when their backend permissions now differ from the frontend
William had written a check on each API call that checked if the backend permissions differed from those that were sent to the frontend on login.

However, once we removed the actual permssions from the AuthenticatedUser in session state - that check was not easy to make.

So now permissions as used by API layer are *always* based on what is actually in the backend database - so people can never actually reach outside their permssions. HOWEVER, what is possible is that the frontend UI thinks that they have permissions that they no longer have - and is therefore showing them parts of the UI _that cannot possibly work_. 

Need to consider how much of a problem this is and what we might do about it (detect, display a toast message?)

### LINKED ISSUE #402: User permissions shown in Edit perms dialog I think comes from UI permission knowledge - not backend
Now that the backend database state is the sole arbiter of permissions - we probably need an API call where the UI asks for the backend permissions of users (or returns the permissions in the GetUsers call?) - rather than using anything known to the frontend. (note: I think it definitely already return the permissions of every user in the GetUsers call - so must be treating the "logged in user" differently)

Currently, I changed my own permissions in the UI - hit Save - then went back to the Edit dialog and the permissions had become unchecked.
(to be clear - the Change permissions had been correctly enacted - it just isn't reflected until I logout and login).

This is related to #401 - which previously immediately forced a login and so hid this.

williamputraintan:
Closes #401 and #402

- Move information stored in cookie (except `csrf` token) to use react context
- Change user permission will trigger re-query on the getOwnUser data and will re-render only component that uses this info

As expected, not re-rendering everything as the top provider state change (see `DAC` in NavBar)
<video src="https://github.com/umccr/elsa-data/assets/61998484/263884b7-a80b-4de0-83f4-a1424553c088" />


application/frontend/src/providers/api-provider.tsx
```diff
@@ -0,0 +1,114 @@
...
+import { useShowAlert } from "./show-alert-provider";
+import axios, { AxiosRequestConfig } from "axios";
+
+export const APIProvider: React.FC<Props> = (props: Props) => {
+  const queryClient = new QueryClient({
+    defaultOptions: {
+      queries: {
+        retry: false,
+      },
+    },
```
williamputraintan, line 16: Proposing to disable any useQuery retry, I don't think retry will eventually change the outcome and it will just take longer to throw the error at the user.

andrewpatto: yep agree with that.. I think there might be some API situations where retry is useful - but probably not us

application/frontend/src/index-router.tsx
```diff
@@ -29,14 +29,14 @@ import { AuditEventsPage } from "./pages/audit-events-dashboard/audit-events-das
...
 import { JobsSubPage } from "./pages/releases/jobs-sub-page/jobs-sub-page";
-import { useUiAllowed } from "./hooks/ui-allowed";
 import { DatasetLayout } from "./layouts/layout-base-dataset";
 import { DacLayout } from "./layouts/layout-base-dac";
 import {
   FEATURE_DEV_TEST_USERS_LOGIN,
   FEATURE_RELEASE_COHORT_CONSTRUCTOR,
   FEATURE_RELEASE_DATA_EGRESS_VIEWER,
 } from "@umccr/elsa-constants";
+import { useEnvRelay } from "./providers/env-relay-provider";
```
DoxasticFox:
```suggestion
```

application/frontend/src/pages/users-dashboard/other-users/other-users.tsx
```diff
@@ -185,22 +171,22 @@ export const OtherUsers: React.FC<Props> = ({ pageSize }) => {
   };
 
   return (
-    <Box heading="Other Users">
+    <Box heading="All Users">
```
DoxasticFox, line 174: Maybe rename this file from `other-users/other-users.tsx` to `all-users/all-users.tsx`

DoxasticFox (approved):

application/frontend/src/pages/users-dashboard/other-users/other-users.tsx
williamputraintan: Yup, changing it! Thanks

mmalenic (approved):

andrewpatto (approved):

PR #434, opened 2023-07-18
Add a summary of data size under the Data Access and Data Location

### LINKED ISSUE #422: Add a summary of data size under the Data Access and Data Location
The cases box can tell us about how many specimens/cases are in the release - but the actual size of the data depends on whether we are releasing files via S3 v reads v variants etc.

So proposing another UI element down in that area which shows that compute - possibly in response to a button press (rather than always)

DoxasticFox:
Closes #422 except for the part where the size gets saved and is viewable by researchers.

https://github.com/umccr/elsa-data/assets/11529449/632702ba-f68e-4aca-8cad-92b2e83fb782

### DIFF
--- a/application/frontend/src/pages/releases/detail/further-restrictions-box.tsx
+++ b/application/frontend/src/pages/releases/detail/further-restrictions-box.tsx
@@ -13,7 +13,6 @@ import { axiosPatchOperationMutationFn } from "../queries";
 import { trpc } from "../../../helpers/trpc";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faFile } from "@fortawesome/free-solid-svg-icons";
-import axios from "axios";
 import { ReleaseSizeType } from "@umccr/elsa-types";
 import { fileSize } from "humanize-plus";
 
@@ -72,6 +71,21 @@ export const FurtherRestrictionsBox: React.FC<Props> = ({
     }
   );
 
+  const [releaseSize, setReleaseSize] = useState<
+    "unknown" | "loading" | ReleaseSizeType
+  >("unknown");
+
+  const releaseSizeQuery = trpc.manifest.getReleaseSize.useQuery(
+    { releaseKey },
+    { enabled: false }
+  );
+
+  const onPressComputeSize = useCallback(async () => {
+    setReleaseSize("loading");
+    const result = await releaseSizeQuery.refetch();
+    setReleaseSize(result.data ? result.data : "unknown");
+  }, [releaseKey]);
+
   const isAllowedCheck = (
     label: ReactNode,
     path:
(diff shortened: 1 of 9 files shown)


application/frontend/src/pages/releases/detail/further-restrictions-box.tsx
```diff
@@ -11,6 +11,48 @@ import {
 import { RhCheckItem, RhChecks } from "../../../components/rh/rh-checks";
 import { axiosPatchOperationMutationFn } from "../queries";
 import { trpc } from "../../../helpers/trpc";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faFile } from "@fortawesome/free-solid-svg-icons";
+import axios from "axios";
+import { ReleaseSizeType } from "@umccr/elsa-types";
+
```
williamputraintan: We have a fileSize formatter function from `humanize-plus` if it helps simplifies this.

`import { fileSize } from "humanize-plus";`

application/frontend/src/pages/releases/detail/further-restrictions-box.tsx
```diff
@@ -69,6 +111,16 @@ export const FurtherRestrictionsBox: React.FC<Props> = ({
...
+  const [releaseSize, setReleaseSize] = useState<
+    "unknown" | "loading" | ReleaseSizeType
+  >("unknown");
+
+  const onPressComputeSize = useCallback(async () => {
+    setReleaseSize("loading");
+    const response = await axios.get(`/api/releases/${releaseKey}/size`);
+    setReleaseSize(response.data);
+  }, [releaseKey]);
+
```
williamputraintan: Wondering if we should have this as a `trpc` instead?

williamputraintan (approved):

application/frontend/src/pages/releases/detail/further-restrictions-box.tsx
mmalenic: Agreed on the tRPC here. I think it could go in `release-router.ts`, or even a new `manifest-router.ts`.

mmalenic (approved):

application/frontend/src/pages/releases/detail/further-restrictions-box.tsx
DoxasticFox: I'm looking at their [examples](https://github.com/HubSpot/humanize#filesize), and the output looks wrong. They've got this:

```typescript
Humanize.fileSize(1024 * 20)
// "20 Kb"
```

"Kb" means "Kilobits". But if the input to `fileSize` is in bytes, it should be Kilo _bytes_, which has an upper case "B" (i.e. KB not Kb). But that's still not right because "kilo-" means 1000, not 1024. Kibi- is the prefix for 1024. So instead of being "20 Kb", `Humanize.fileSize(1024 * 20)` should be "20 KiB" or "20.48 KB".

DoxasticFox: Oh, actually their code gets the capitalisation of the "B" right. They still don't respect the difference between "K-" and "Ki-", but, eh, that only makes the number slightly wrong.

DoxasticFox: ...Except I also made the same mistake in my code.

andrewpatto (approved):

PR #436, opened 2023-07-20
Feature/replace cases query with edgeql

### LINKED ISSUE #428: Rewrite the cases query as EdgeQl
The GetCases query was the first one we wrote against edgedb - but is the most in need of an optimisation/rewrite as edgeql.

We will be able to pick up a heap of info we currently re-compute via two or three queries - all in one query.

andrewpatto:


application/backend/dbschema/queries/release-selection/releaseSelectionGetCases.edgeql
```diff
@@ -0,0 +1,86 @@
...
+  # in order that the case is returned
+  paramQuery := <optional str>$query,
+
+  # if present, use as a paging offset
+  paramOffset := <optional int64>$offset,
+
+  # if present, use as a paging limit
+  paramLimit := <optional int64>$limit,
+
+  #######
```
andrewpatto, line 22: Experimenting with a possible pattern for how to document the parameters to our EdgeQl..

I am not a fan otherwise of how edgedb requires the specification of the type at every location a parameter to used (not very DRY). This pattern allows them to be asserted once at the start - and also creates a pattern by which the params can be documented

andrewpatto:
This is just a replacement of the very first typescript edge query we wrote - where it literally made the query twice just to retrieve the count.

Replaced with a much saner pure edgeql version.

I have no idea what the performance difference is - because the new query crashes the Edge query analyzer. It certainly doesn't feel any slower locally (nor any faster tbh). Will find out when deployed I guess.

I think even if slower somehow - we should be looking to optimise this function rather than reverting back to the previous approach.

williamputraintan (approved):

application/backend/dbschema/queries/release-selection/releaseSelectionGetCases.edgeql
mmalenic: I like this style - I think it makes it neater. Not sure how I feel about capitalising the key words, but it does make them stand out more.

mmalenic (approved):

PR #438, opened 2023-07-27
Egress Summary Warning not part of Activated list

### LINKED ISSUE #427: Consider the data egress summary logic across multiple activations
The data egress for the non-summary is clear - show only the records that we have synced over to our db.

For the summary - it gets a list of the files from the manifest to summarise against.

Consider what might happen if the release is deactivated - and then reactivated with 5 less files. Should we do some UI to show these files still in the summary but grayed out?

Also, what is the effect on the summary of the release being deactivated entirely (which possibly shouldn't affect the egress summary as that is activity that has already occurred)

williamputraintan:
Fix #427

Screenshot 2023-07-28 at 13 26 58

### DIFF
--- a/application/backend/dbschema/queries/release-data-egress/getReleaseDataEgressSummary.edgeql
+++ b/application/backend/dbschema/queries/release-data-egress/getReleaseDataEgressSummary.edgeql
@@ -3,8 +3,8 @@
 with
 
   # There are 2 scenarios where the list of files could come from:
-  # (1) From a manifest file that WAS activated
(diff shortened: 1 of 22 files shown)


andrewpatto:
Is it possible to add a test showcasing this query - both to show this feature - but also to flesh out the testing around the activation stuff?

Could be a direct db-test that just exercises the query in different circumstnaces (no activation, 1 activation, 2 activations)

andrewpatto (approved):

PR #445, opened 2023-08-01
refactor: audit event query

### LINKED ISSUE #326: Audit entry backend should use unified query
There is now a unified query mechanism for audit events.. BUT - it cannot do "ORDER BY" with the same granularity as the typescript version. In particular - there is no way to pass in as a param the ASC or DESC aspect of an order..

Might need a discussion in the EdgeDb boards - cos maybe I am just missing something..

So once we have a single function - we can replace a lot of the backend Audit entry service methods.

mmalenic:
Closes #326

### Changes
* Add long if else for sorting audit queries.
* Remove client library audit log queries.
* Add some tests.


DoxasticFox (approved):

application/backend/src/business/services/user-service.ts
```diff
@@ -342,34 +349,23 @@ export class UserService {
     whoDisplayName: string
   ) {
     await client.transaction(async (tx) => {
-      const release = await e
-        .select(e.release.Release, (r) => ({
-          releaseKey: true,
-          filter_single: { id: r.id },
-        }))
-        .run(tx);
+      const { releaseKey } = await releaseGetKeyByDbId(tx, {
```
williamputraintan: Perhaps we could pass the releaseKey from the parent function so we could move from relying the releaseUUID. I guess could do this on seperate PR.

application/backend/dbschema/queries/audit-event/auditEventGetSomeByUser.edgeql
```diff
@@ -77,17 +83,44 @@ select {
...
+      <str>.details if orderByProperty = "details" and <str>orderAscending = "false" else
+      <str>.occurredDateTime if orderByProperty = "occurredDateTime" and <str>orderAscending = "false" else
+      <str>.occurredDuration if orderByProperty = "occurredDuration" and <str>orderAscending = "false" else
+      <str>.outcome if orderByProperty = "outcome" and <str>orderAscending = "false" else
+      <str>.recordedDateTime if orderByProperty = "recordedDateTime" and <str>orderAscending = "false" else
+      <str>.inProgress if orderByProperty = "inProgress" and <str>orderAscending = "false" else
+      <str>.whoId if orderByProperty = "whoId" and <str>orderAscending = "false" else
+      <str>.whoDisplayName if orderByProperty = "whoDisplayName" and <str>orderAscending = "false" else
+      <str>.occurredDateTime
+    ) desc
```
andrewpatto, line 123: yikes! (totally understand that this is the only solution to this but yikes!)

andrewpatto: should orderAscending be a str? Couldn't it be an actual boolean param? I mean it looks like it starts out as a boolean when it comes into the query.

I understand why the `.updatedDateTime` needs to be cast to string to be a compatible orderable - but `orderAscending` is just being used as a boolean operator in an expression

application/backend/dbschema/queries/audit-event/auditEventUserAddedToRelease.edgeql
```diff
@@ -0,0 +1,35 @@
+# add a user audit event when adding a user to a release.
```
andrewpatto: Given the general pattern of this - wouldn't it be better to parametrise the description and details - and make one single query for adding to userAuditEvent? (rather than embedding very specific strings like "add user to release" in the query)?

application/backend/dbschema/queries/audit-event/auditEventUserPermissionsChanged.edgeql
```diff
@@ -0,0 +1,29 @@
+# add a user audit event when adding a user to a release.
```
andrewpatto: i.e. parametrise this pattern and remove this query

application/backend/src/business/services/user-service.ts
```diff
@@ -166,24 +167,30 @@ export class UserService {
...
+      const targetUser = await userGetBySubjectId(tx, {
+        subjectId: targetSubjectId,
+      });
+
+      if (targetUser === undefined || targetUser === null) {
+        throw new NonExistentUser(targetSubjectId);
+      }
+
+      await userUpdatePermissions(tx, {
+        userDbId: targetUser.id,
```
williamputraintan: I am also wondering if we should also retain the filter based on subjectId instead of dbId

williamputraintan (approved):

application/backend/src/business/services/user-service.ts
mmalenic: Yes, good call.

application/backend/dbschema/queries/audit-event/auditEventUserAddedToRelease.edgeql
mmalenic: Yeah, I think so too.

application/backend/dbschema/queries/audit-event/auditEventGetSomeByUser.edgeql
mmalenic: Yes, for some reason I thought all elements in the if else had to be cast to the same type, but I think that's only for the return type.

application/backend/src/business/services/dacs/rems-service.ts
```diff
@@ -167,22 +162,23 @@ export class RemsService {
...
-          lastUpdatedSubjectId: user.subjectId,
-          applicationDacIdentifier: e.tuple({
-            system: remsUrl,
-            value: e.str(remsId.toString()),
-          }),
-          applicationDacTitle:
-            application["application/description"] || "Untitled in REMS",
-          applicationDacDetails: `
+        .select(
+          e.insert(e.release.Release, {
```
mmalenic, line 166: Additional select statement.

application/backend/src/business/services/user-service.ts
mmalenic: I've lifted up the `releaseKey` and removed `releaseUUID`, however at some point it stops at an `e.insert(e.release.Release, ...` which only returns the database id, so an additional select statement is needed (see [here](https://github.com/umccr/elsa-data/pull/445#discussion_r1283759123)). I think when these queries get converted to EdgeQL that can be optimised too.

andrewpatto (approved):

PR #452, opened 2023-08-17
Add a bulk select/deselect using text files to the Cases box

### LINKED ISSUE #439: Add a bulk select/deselect using text files to the Cases box
The cohort constructor will be the preferred mechanism to build lists of selected patients - but in the interim - we need a mechanism by which AG staff can use their existing techniques. This involves them building a CSV (TSV?) in excel with existing manual matching (doesn't matter as it is just one column I presume) of ids that they want to release.

We should allow them to drag this onto somewhere in the cases Box - and have all the matches specimens/patients be selected.

DoxasticFox:
Closes #439


application/backend/dbschema/queries/release/releaseGetSpecimensByDbIdsAndExternalIdentifiers.edgeql
```diff
@@ -0,0 +1,77 @@
+# A query to find specimen ids for a release - but respecting the tree structure
```
andrewpatto, line 1: I've added a new query for comparison - more idiomatic edgeql (less FOR loops)

application/backend/src/business/services/releases/release-selection-service.ts
```diff
@@ -322,50 +329,45 @@ export class ReleaseSelectionService extends ReleaseBaseService {
...
-        const crossLinkedSpecimenIds =
-          specimenDatasetLinks && specimenDatasetLinks.crossLinked
-            ? specimenDatasetLinks.crossLinked.map((scl) => scl.id)
-            : [];
-        const invalidSpecimenIds: string[] = [];
-
-        const recognisedSpecimenIds = new Set(
-          validSpecimenIds.concat(crossLinkedSpecimenIds)
-        );
+        // @Christian for finishing.. just matching to the new query or altering the new query to do what you want here
```
andrewpatto: I only just tied it together to see if it worked - but will leave all the rest of the plumbing

application/frontend/src/components/csv-dropzone.tsx
```diff
@@ -0,0 +1,104 @@
...
+    onError: _onDropError,
+    accept: { "text/csv": [] },
+    // I don't think we are prepared in a UI sense to handle multiple CSVs so lets not allow
+    multiple: false,
+  });
+
+  return (
+    <>
+      <div {...getRootProps()} className="flex flex-col items-start gap-6">
+        <div className="flex flex-col items-center justify-center rounded-2xl border-2 border-dashed border-slate-400 bg-gray-200 p-4 hover:bg-gray-100">
```
andrewpatto, line 103: I took out some of the padding and min-widths.. would be happy to have them back in but maybe as passed in props or something? (I wanted the dropzones on the cases to be smaller)

application/frontend/src/pages/releases/detail/cases-box/cases-box.tsx
```diff
@@ -338,37 +338,45 @@ export const CasesBox: React.FC<Props> = ({
...
+                        {currentSelectedSpecimens} specimen
+                        {currentSelectedSpecimens !== 1 && "s"} in total
+                        selected
+                      </span>
+                    </div>
+                    <BulkSelectionDiv
+                      releaseKey={releaseKey}
+                      isAllowEdit={isAllowEdit}
+                      releaseIsActivated={releaseIsActivated}
+                    />
```
andrewpatto, line 398: I moved this into the actual cases box (sorry it wasn't clear from my rough sketch that was what I meant).

andrewpatto:
Just some notes - not sure what the status was so don't know if I've broken these or they were still WIP
- when the release is activated/editing is not allowed - the dropzones still allows CSV to be dropped on them (would be better if they disabled)
- I couldn't get the format of a CSV working to test this out - but that might just have been my CSV (single column I presume?)

application/frontend/src/pages/releases/detail/cases-box/cases-box.tsx
DoxasticFox: All good. I should have understood from the drawing; it makes much more sense for the dropzones to be part of the cases box.

application/backend/dbschema/queries/release/releaseGetSpecimensByDbIdsAndExternalIdentifiers.edgeql
DoxasticFox: Thanks. That was very, very helpful

application/frontend/src/components/csv-dropzone.tsx
```diff
@@ -0,0 +1,110 @@
...
+        if (results.errors.length > 0) {
+          return reject(results.errors);
+        } else {
+          return resolve(results.data);
+        }
+      },
+      error: (error) => {
+        return reject([error]);
+      },
+      delimiter: ",",
```
DoxasticFox, line 43: @andrewpatto This line should address this feedback:

>I couldn't get the format of a CSV working to test this out - but that might just have been my CSV (single column I presume?)

application/backend/src/api/routes/internal-trpc/release/release-router.ts
```diff
@@ -32,7 +32,9 @@ const casesQuerySchema = inputReleaseKeySingle
 const specimensMutateSchema = inputReleaseKeySingle.merge(
   z.object({
     op: z.literal("remove").or(z.literal("add")),
-    value: z.array(z.string()),
+    dbIds: z.array(z.string()),
+    externalIdentifierValues: z.array(z.string()),
+    selectAll: z.boolean(),
```
andrewpatto: can we perhaps do this as an either or.. either dbIds/externalIds OR selectAll.. I presume if select all is set then the other fields are ignored. (thinking then - it probably should be selectAll: z.literal(true). As selectAll true is the only operation with that field that makes any sense?

andrewpatto (approved):

andrewpatto:
Looks good.

Possibly extend the Cypress test that involves selecting cases/nodes - could we also simulate dropping a CSV on? (not sure how hard this is in cypress - if too hard happy to skip)

williamputraintan (approved):

DoxasticFox:
>Possibly extend the Cypress test that involves selecting cases/nodes

Even though the backend has that ability now, I didn't wire-up the front end for selecting patients/cases. I'd rather do it in a separate PR, if you're okay with that; +800 LOC is already ginormous IMO. (**Edit:** Issue to track that work: https://github.com/umccr/elsa-data/issues/457)

>could we also simulate dropping a CSV on? (not sure how hard this is in cypress - if too hard happy to skip)

:heavy_check_mark:

PR #454, opened 2023-08-21
Stricter login

### LINKED ISSUE #447: Logging with a IdP that provides no email halts the login (correctly) but redirects to an invalid page
When people try to login with ORCID - it does not provide an email address - which is fundamental to how Elsa works.

So we've made the decision that the login should abort - but we currently redirect to a page that doesn't exist.

Need to make a "public" page that explains this and allows them to naviagate back to the main login page

(we could consider separately about whether we can convince ORCID to give us email claims - but not in this issue)

### LINKED ISSUE #451: Reduce surface by only allowing User or PotentialUser to login
At the moment we allow anyone who can log in via CILogon/AAF to login - albeit they come in as a user who has no permission to do anything.

HOWEVER - this does mean that essentially anyone on the internet can obtain a JWT - that could be used to explore API weaknesses.

Instead we could consider not allowing a login *unless* the email *already* exists from the set

User
PotentialUser

or the subject id exists from the set

Superadmin

The only issue would be that we need a way of bootstrapping AG admin staff.

andrewpatto:
Only allow those who already exist in the db or are a super admin
Also added CMD to add new users


andrewpatto:
Whilst here also tightened up the OIDC flow to use state and nonce values
Also, OIDC errors can redirect to frontend pages specific for different errors

andrewpatto:
I would have liked to add an E2E test - but we would need a external OIDC provider to run the tests through..

andrewpatto:
Closes #451 
Closes #447

DoxasticFox (approved):

Issue #433, opened 2023-07-17
Investigate SOCI indexer for Elsa images


andrewpatto:
https://aws.amazon.com/blogs/aws/aws-fargate-enables-faster-container-startup-using-seekable-oci/

andrewpatto:
Have added an installer for the soci binary to the github actions.. 

But the docker build push is all one action - so not sure how to fit the invoke of soci in

andrewpatto:
Progressed it a bit - but the soci github action is like one week old - so everything has rough edges. Revisit in a few months

andrewpatto:
Given the eventual deployed image (for AWS) it rebuilt by our CDK - this might also be of interest.

https://github.com/aws/aws-cdk/issues/26413

PR #456, opened 2023-08-22
attempt at soci build

andrewpatto:


andrewpatto:
I know this is kind of pointless to merge because it doesn't actually do anything - but the pattern is sound and does work.. we just need soci/ghcr etc to progress the support a bit. So am suggesting a merge and then revisit in a few months - rather than hanging out on a branch that will get forgotten

andrewpatto:
Progressing #433

mmalenic (approved):

PR #465, opened 2023-08-30
Invite Potential User through API/UI

### LINKED ISSUE #459: A UI that allows super admins to invite users into the instance
Now that only people listed in Users or PotentialUsers can log in - we need a way that the super admins can invite
people into the instance - without making them part of a release.

So a simple UI - with an email address box (and perhaps the initial permissions to give) - that only the super admins
can see/use. And it hits an API that adds that user into PotentialUser table (must already check for the existence of the user)

(I've built a rudimentary way of doing this with the Elsa CLI tool - but a UI would be better long term)

### LINKED ISSUE #460: Make potential user permissions booleans transfer to new user
I put some boolean fields in potentialuser to mirror the boolean in the real user. The idea being that if they are set in potential user - a person can automatically be given permissions.

However, there was no mechanism to actually set the boolean - so I didn't implement the transfer.

Now with #459 on the cards - we *can* set the potential user booleans. So we should do the same when transferring a potential to real user.

I imagine this is like 5 extra lines of edgeql

williamputraintan:
Feature #459
Fix #460


williamputraintan:
So I have changed the permission naming in the potentialUser schema removing the `future` word. This is mainly to reuse the component/keywords in FE. I think having the same property name for the active or potential users should not cause confusion. Happy to revert back ( to use `future` in permission) if it makes more sense.

application/backend/dbschema/permission.esdl
```diff
@@ -45,15 +45,15 @@ module permission {
         # whether this user will be given extra rights when they first login
```
williamputraintan, line 45: Changing the permission naming in the potentialUser schema removing the `future` word. This is mainly to reuse the component/keywords in FE. Having the same property name for the active or potential users should not cause confusion I think. Happy to revert back ( to use `future` in permission) if it makes more sense.

williamputraintan:
- Add PU
<video src="https://github.com/elsa-data/elsa-data/assets/61998484/58d40595-0060-45f3-adab-1c5335222731" />

williamputraintan:
- Edit/remove PU
<video src="https://github.com/elsa-data/elsa-data/assets/61998484/76d9c77b-c45b-4a73-b225-38f919b23823" />

application/frontend/src/pages/users-dashboard/all-users/components/invite-potential-user.tsx
```diff
@@ -0,0 +1,212 @@
...
+            <>
+              <button
+                type="button"
+                disabled={
+                  isLoadingMutate ||
+                  !isPotentialEmailValid ||
+                  invitePotentialUser.isSuccess ||
+                  !isEditingAllowed
+                }
+                className="inline-flex w-full items-center justify-center rounded-md border border-transparent bg-red-600 px-4 py-2 text-base font-medium text-white shadow-sm hover:bg-red-700 focus:outline-none focus:ring-2 focus:ring-red-500 focus:ring-offset-2 disabled:opacity-50 sm:ml-3 sm:w-auto sm:text-sm"
```
andrewpatto: can't we use some common daisy button classes?

andrewpatto: we'll never keep all the focus ring opacities etc synchronised across the whole code base.. if we do want specialist buttons - then we need to make some master button CSS styles

andrewpatto (approved):

andrewpatto:
we are also going to do a complete edgedb migration reset
(though happy if you just follow this up with a new PR that does just that)

application/frontend/src/pages/users-dashboard/all-users/components/invite-potential-user.tsx
williamputraintan: Yeah, I was reusing the same css as what we currently have in the beginning. But I guess we could start moving to the daisyUI instead.

williamputraintan:
> we are also going to do a complete edgedb migration reset
(though happy if you just follow this up with a new PR that does just that)

Ah right, will do this in this PR.

application/frontend/src/pages/users-dashboard/all-users/all-users.tsx
```diff
@@ -55,141 +50,48 @@ const permissionIconProperties: {
...
+        >
+          Active
+        </a>
+        <a
+          onClick={() => setUserTableView(UserTableType.POTENTIAL)}
+          className={classNames("tab-lifted tab", {
+            "tab-active": userTableView == UserTableType.POTENTIAL,
+          })}
+        >
+          Potential
```
mmalenic: Would it be better to just call this "Invite"? I feel like the users of Elsa won't have a concept of a `PotentialUser`, and if this tab is the only way to invite other users I think that would make it clearer.

mmalenic (approved):

application/frontend/src/pages/users-dashboard/all-users/all-users.tsx
williamputraintan: Fair enough, I just followed our schema naming so we don't confused. `invite` sounds good to me too.

PR #467, opened 2023-09-01
Feature/rems reborn

andrewpatto:
Not in any way finished - but restored REMS so at least it can now be demoed


application/backend/src/api/routes/internal-trpc/dac-router.ts
```diff
@@ -11,6 +11,18 @@ export const dacRouter = router({
 
     return ctx.dacService.getConfigured(user);
   }),
+  detectNewQuery: internalProcedure
```
andrewpatto, line 14: this is a bit messy - added a duplicate DAC query because the pattern between REMS and AG is not quite the same.. to be revisited down the track

mmalenic (approved):

PR #470, opened 2023-09-06
Only grey-out case row

### LINKED ISSUE #469: Whole case select/unselect grays out the entire page
When changing the checkbox of individual samples - only the single case row is "grayed out" during the operation.

However, when changing the checkbox of a case - the entire page of cases "grays out" during the operation. Wondering if it is possible to get it to behave the same as the sample checkbox case?

DoxasticFox:
Closes https://github.com/elsa-data/elsa-data/issues/469

### DIFF
--- a/application/frontend/src/pages/releases/detail/cases-box/cases-box.tsx
+++ b/application/frontend/src/pages/releases/detail/cases-box/cases-box.tsx
@@ -142,7 +142,7 @@ export const CasesBox: React.FC<Props> = ({
     }
   );
 
-  const casesQueryData = casesQuery.data?.data;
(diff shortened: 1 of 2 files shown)


andrewpatto:
So you've dropped the operation one level down in the react component hierarchy?

andrewpatto (approved):

DoxasticFox:
@andrewpatto Yep. Sounds good?

Demo I should've added to the OP:

https://github.com/elsa-data/elsa-data/assets/11529449/fcb0f9ba-3643-4fd0-885a-daa861a3ae21

andrewpatto:
Yep sounds like a good approach. Looks good.

andrewpatto:
I'm about to make a proper release - so will merge on your behalf now

PR #472, opened 2023-09-11
Feature/142 access point improve

### LINKED ISSUE #142: Split Access Point sharing CloudFormation scripts if they get too large
Currently, the S3 access point mechanism copes with sourcing data from multiple buckets by creating Nested cloud formation stacks.

However, CloudFormation has a 1MB limit on cfn scripts - and we do not currently check for that. If there are more than (estimate 100) files added into a policy - then the entire cfn script will overflow the size.

What we need to do is to detect this situation as we build the scripts - and just overflow into another nested access point.
(which should be relatively straight forward as the mechanism for multiple stacks is already in place).

### LINKED ISSUE #405: Consider what data should be released in the TSV manifest
The TSV manifest is always available to everyone involved in a release - and is meant to pretty much just be an easy way to download what is present in the Cases box (that they can see with their eyes). So currently it lets them download some things like the S3 URL path (which in itself is not dangerous) - but I wonder if we should have a smaller subset for this manifest.

andrewpatto:
Changed the templating mechanism to allow hundred of objects (will overflow eventually - but more like 1000+)
Made the access point mechanism use regular manifest mechanisms


application/backend/.env.aws
```diff
@@ -14,4 +14,4 @@ ELSA_DATA_VERSION=local-dev-aws
 ELSA_DATA_BUILT=sometime
 ELSA_DATA_REVISION=0000000000000000000000000000000000000000
 
-ELSA_DATA_META_CONFIG_SOURCES="file('base') file('umccr-garvan-dev-super-admins') file('dev-localhost') file('datasets') file('sharers') aws-secret('ElsaDataLocalhost')"
+ELSA_DATA_META_CONFIG_SOURCES="file('base') file('umccr-garvan-dev-super-admins') file('dev-localhost') file('disable-mock-aws') file('datasets') file('sharers') aws-secret('ElsaDataLocalhost')"
```
andrewpatto, line 17: The "aws" env is for developing locally but where it actually hits AWS services - so we need to switch off mocking

application/backend/src/business/services/aws/aws-cloudtrail-lake-service.ts
```diff
@@ -289,13 +289,14 @@ export class AwsCloudTrailLakeService {
             );
           }
         } else if (method == CloudTrailQueryType.S3AccessPoint) {
-          const bucketNameMap = (
-            await this.awsAccessPointService.getInstalledAccessPointResources(
+          const map =
```
andrewpatto, line 292: @williamputraintan I think I have converted this logic correctly but we possibly just need to do some experiements to confirm

application/backend/src/business/services/manifests/htsget/manifest-htsget-service.ts
```diff
@@ -74,6 +83,92 @@ export abstract class ManifestHtsgetService {
...
+    const htsgetManifest = await this.getActiveHtsgetManifest(releaseKey);
+
+    if (!htsgetManifest) throw new Error("needs an active htsget manifest");
+
+    const tsvObjects: ManifestBucketKeyObjectType[] = [];
+
+    // TODO - better filling in of the TSV objects structure
+    //        might need cross referencing from the master manifest etc
+
+    for (const [key, r] of Object.entries(htsgetManifest.reads ?? {})) {
```
andrewpatto, line 121: @mmalenic I've just done the skeleton here. You can follow up with a proper PR. It is hooked into the frontend.

application/backend/src/business/services/manifests/manifest-bucket-key-types.ts
```diff
@@ -23,11 +27,11 @@ export const ManifestBucketKeyObjectSchema = Type.Object({
...
-assert(
-  JSON.stringify(
-    [...Object.keys(ManifestBucketKeyObjectSchema.properties)].sort()
-  ) === JSON.stringify([...ObjectStoreRecordKey].sort())
-);
+//assert(
+//  JSON.stringify(
+//    [...Object.keys(ManifestBucketKeyObjectSchema.properties)].sort()
+//  ) === JSON.stringify([...ObjectStoreRecordKey].sort())
+//);
```
DoxasticFox: I remember wanting to do this with a typecheck, but I couldn't, so I gave up and did this janky thing in frustration :joy: . I think as a general rule, it's better to delete stuff completely than leave it in as a comment, so I vote to delete this

```suggestion
```

DoxasticFox (approved):

application/backend/src/business/services/manifests/manifest-bucket-key-types.ts
andrewpatto: sorry yes! meant to chase up with you and ask about it.. it started failing at some point during dev.

andrewpatto: I don't remember what I did!

andrewpatto: I think possibly I was bumping packages and it didn't survive the upgrade to the new typescript. Is it ok to delete - or should I try to get it working with the new typescript?

DoxasticFox: Yeah, I don't remember anything breaking without it. I just wanted to enforce that the keys of `ManifestBucketKeyObjectSchema` were the elements of `ObjectStoreRecordKey`

DoxasticFox: I assume the tests would indirectly check that's true

application/backend/src/business/services/manifests/htsget/manifest-htsget-service.ts
mmalenic: Okay, the idea behind this is that it's a different structure to what htsget uses when calling Elsa in the `publishHtsgetManifest` function? I.e. it's a per-object flattened representation?

application/frontend/src/pages/releases/detail/access-box/manifest-form.tsx
```diff
@@ -5,69 +5,36 @@ import { TsvColumnCheck } from "../../../../components/access-box";
...
-      </form>
+      <TsvDownloadDiv
+        actionUrl={`/api/releases/${releaseKey}/tsv-manifest-plaintext`}
+        releaseActivated={!!releaseData.activation}
+        fieldsToExclude={[
+          "objectStoreSigned",
+          "objectStoreBucket",
+          "objectStoreKey",
+          "objectStoreUrl",
+        ]}
```
mmalenic, line 36: Does this already take into account filtering based on the activated "s3", "gcp" and "r2" storage locations?

mmalenic (approved):

application/frontend/src/pages/releases/detail/access-box/manifest-form.tsx
andrewpatto: Yes - the Master manifest is the original source of truth - and should only have in it things that are allowed to be released by the rules of the admins (including by storage location)

All the other manifests are just transforms of that master data.

application/backend/src/business/services/manifests/htsget/manifest-htsget-service.ts
andrewpatto: yes - it just a simple structure they can import into excel or R etc
(may be useful, may not be useful - but it adds some symmetry to the other access mechanisms that all have something similar)

PR #485, opened 2023-10-04
Feature/dataset sync only by admin

andrewpatto:
Moving dataset sync functions to only be available via CMD entrypoint.
This will allow us to split the AWS permissions between the Web and Cmd - so that the Web does not have *any* access to any genomic files. Which lessens any breach surface.

Also, had a strange regression in just the test suite - to do with cilogon now rejected oidc discovery. So have had to introduce some workarounds for that.

### DIFF
--- a/application/backend/config/dev-localhost.json5
+++ b/application/backend/config/dev-localhost.json5
@@ -3,7 +3,7 @@
   // should then overwrite these empty values
   // in the absence of setting these elsewhere - at least the system will start up - albeit with login problems!
   oidc: {
-    issuerUrl: "https://test.cilogon.org",
+    issuerUrl: "",
     clientId: "aclientid",
     clientSecret: "abc",
   },
--- a/application/backend/src/api/api-auth-routes.ts
+++ b/application/backend/src/api/api-auth-routes.ts
@@ -31,8 +31,15 @@ import {
 const FILENAME_FOR_LOGGING = "api-auth-routes";
 
 function createClient(settings: ElsaSettings, redirectUri: string) {
-  if (!settings.oidc || !settings.oidc.issuer)
-    throw new Error("Cannot establish OIDC login without OIDC settings");
+  // we allow the absence of settings to silently go through - but will display
+  // a proper exception if the undefined BaseClient is attempted to be used in practice
+  if (
+    !settings.oidc ||
+    !settings.oidc.issuer ||
+    !settings.oidc.clientId ||
+    !settings.oidc.clientSecret
+  )
+    return undefined;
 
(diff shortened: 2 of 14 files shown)


application/backend/config/dev-localhost.json5
```diff
@@ -3,7 +3,7 @@
   // should then overwrite these empty values
   // in the absence of setting these elsewhere - at least the system will start up - albeit with login problems!
   oidc: {
-    issuerUrl: "https://test.cilogon.org",
+    issuerUrl: "",
```
andrewpatto, line 6: issuer discovery now seems to return "unauthorized" from the test.cilogon host.. we really should only have been doing OIDC *when* we actually want to do an OIDC flow, and not in dev/localhost test user anyhow.

application/backend/src/api/api-auth-routes.ts
```diff
@@ -68,6 +75,11 @@ export const apiAuthRoutes = async (
   // it constructs the appropriate auth url (as known to the backend only)
   // and sends it back to the client as a redirect to start the oidc flow
   fastify.post("/login", async (request, reply) => {
+    if (!client)
+      throw new Error(
+        "OIDC client not configured so cannot proceed with an OIDC login flow"
+      );
+
```
andrewpatto, line 82: The discovery of the lack of OIDC settings is now deferred till an OIDC flow is actually attempted.

application/backend/src/api/routes/internal-trpc/dataset-router.ts
```diff
@@ -40,30 +40,6 @@ export const datasetRouter = router({
       const { datasetUri, includeDeletedFile = false } = input;
       return await ctx.datasetService.get(user, datasetUri, includeDeletedFile);
     }),
-  updateDataset: internalProcedure
```
andrewpatto, line 43: removed the API entrypoint for updating datasets

application/backend/src/business/services/australian-genomics/s3-index-import-service.ts
```diff
@@ -1020,11 +1012,11 @@ export class S3IndexApplicationService {
...
-      datasetUri,
-      now,
-      this.edgeDbClient
-    );
+    // await this.auditLogService.insertSyncDatasetAuditEvent(
+    //  user,
+    //  datasetUri,
+    //  now,
+    //  this.edgeDbClient
+    // );
```
andrewpatto, line 1020: once the sync dataset is established I will rethink the audit events - at the moment just disabling

williamputraintan (approved):

mmalenic (approved):

PR #491, opened 2023-10-09
Removed some exceptions when making a manifest just for stats

andrewpatto:


andrewpatto:
Computing "stats" - but where all the reads/variants we not yet set up - was returning "0 bytes / 0 files" because of a different exception

DoxasticFox (approved):

application/backend/src/business/services/manifests/manifest-master-helper.ts
```diff
@@ -14,15 +17,19 @@ const R2_PREFIX = "r2://";
...
  *
  * @param manifest the input manifest to be transformed
+ * @param throwExceptions exceptions are thrown in the process of creating
+ *                        an actual manifest - but there are cases where we want
+ *                        stats from the manifest and do not want to throw exceptions
  */
 export async function transformDbManifestToMasterManifest(
-  manifest: ManifestMasterType
+  manifest: ManifestMasterType,
+  throwExceptions: boolean = true
```
mmalenic, line 26: Would it be better to have two functions here instead? One of them could construct the manifest without throwing, and another checks all the throwable conditions.

mmalenic (approved):

application/backend/src/business/services/manifests/manifest-master-helper.ts
andrewpatto: yeah agreed - we possibly need to revisit all the manifest stuff - and add some better layering (and maybe some splitting). I suggest leaving for the moment as this was a minimal change more for a bug

PR #492, opened 2023-10-09
Added config setting to allow admins to download objects

### LINKED ISSUE #489: Allow download mechanisms to be viewed by admins (based on configuration setting)
Whilst I don't think it is necessary in production - for demonstration it is useful if administrators can also download manifests etc.

At the moment that entire bit of the UI *only* appears for researchers.

Add a configuration that enables the display of this UI to admins

andrewpatto:


application/backend/src/business/services/permission-service.ts
```diff
@@ -0,0 +1,34 @@
...
+
+   * @returns
+   */
+  public canAccessData(userRole: UserRoleInRelease): boolean {
+    return (
+      userRole === "Manager" ||
+      userRole === "Member" ||
+      (!!this.settings.permission?.releaseAdministratorsCanAlsoAccessData &&
+        userRole === "Administrator")
+    );
```
andrewpatto, line 35: started on a pattern by which we can centralise any repeated "permission logic" (that isn't in edgeql)

mmalenic (approved):

application/backend/src/business/services/permission-service.ts
mmalenic: I like this idea - makes the repeated logic clearer I think.

PR #493, opened 2023-10-10
Fix logic for 'Audit Events' spinner

### LINKED ISSUE #490: The "loading" behaviour for the releases and datasets table is different to the audit table
The release and datasets table both display a "snowflake .. loading..." (with no table at all) when they are doing an initial/slow load. Some level of feedback is important - due to the fact our db is a bit slow with cold starts.

On the other hand, the audit events table shows the table header - but with *no* "activity" animation on the same initial/slow load. So this needs to be fixed.

But I also think we need to unify the approach across all the React tables - but could go either way

- audit event table is made to display blank with "snowflake loading" (i.e. behave identical to release and datasets)
- audit event table shows some form of "snowflake.. loading" *inside* the otherwise empty table structure - and then we convert release and datasets to behave the same

Open to suggestions..

DoxasticFox:
Closes https://github.com/elsa-data/elsa-data/issues/490

### DIFF
--- a/application/frontend/src/components/audit-event/audit-event-table.tsx
+++ b/application/frontend/src/components/audit-event/audit-event-table.tsx
@@ -166,8 +166,6 @@ export const AuditEventTable = ({
     manualSorting: true,
   });
 
-  if (isAnyAuditEventQueryFetching(dataQueries)) return <IsLoadingDiv />;
-
   // TODO Search and filtering functionality, refresh button, download audit log button
   return (
     <Box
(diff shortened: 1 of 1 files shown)


application/frontend/src/components/audit-event/audit-event-table.tsx
```diff
@@ -158,8 +166,6 @@ export const AuditEventTable = ({
     manualSorting: true,
   });
 
-  if (dataQueries.isFetching) return <IsLoadingDiv />;
```
DoxasticFox, line 161: `dataQueries.isFetching` was always falsey because `isFetching` doesn't exist on `dataQueries`. `dataQueries` is the object returned by `useAllAuditEventQueries`. The typechecker could have caught this if the return type of `useAllAuditEventQueries` wasn't explicitly set to `{ [key: string]: UseQueryResult<AuditEventType[]> }`. So this PR is mostly the result of removing that return type and fixing the errors which popped up as a result.

I also moved the `IsLoadingDiv` component into the `Box` component, for consistency with the "Releases" and "Datasets" tables.

andrewpatto: Looks good

andrewpatto (approved):

PR #496, opened 2023-10-11
Redirect from `/login` to `/releases` when already logged in

### LINKED ISSUE #488: Visting the login URL when logged in still displays the login button
We correctly bounce people through to `/releases` when we log in.

Also, if we hit the `/` page - we bounce people to the correct logged in `/releases`.

However - if we are logged in - and visit `/login` - it still shows a Login button. I think instead it should either bounce - or display other buttons (Login grayed out? - and some text about how to logout?)

Screenshot 2023-10-07 at 11 06 01 am

DoxasticFox:
Closes https://github.com/elsa-data/elsa-data/issues/488


application/frontend/src/pages/login-page.tsx
```diff
@@ -1,12 +1,20 @@
...
   showDevTestLogin: boolean;
 };
 
-export const LoginPage: React.FC<Props> = ({ showDevTestLogin }) => {
+export const LoginPageOrRedirect: React.FC<Props> = ({ showDevTestLogin }) => {
   const envRelay = useEnvRelay();
+  const userObject = useLoggedInUser();
+
+  if (userObject) {
+    return <Navigate to={"/releases"} replace />;
```
andrewpatto: I presume we might have "/releases" as the target of a redirect else where.. just so we don't miss this if we potentially refactored - can we move them all to a common const string declaration somewhere? In one of the routes ts?

andrewpatto (approved):

PR #499, opened 2023-10-12
(Un)mount components less frequently in `InformationBox`

### LINKED ISSUE #479: Job running UI has some visual artifacts due to refresh
https://github.com/elsa-data/elsa-data/assets/1973827/09fed652-58a9-4657-bd6e-a98f465052dc

Presumably because of the way we do the timed refresh of the status - it causes the CSS animations to continuously trigger.

I presume this just needs a bit of a restructure of the React tree..

DoxasticFox:
Closes https://github.com/elsa-data/elsa-data/issues/479

### DIFF
--- a/application/frontend/src/pages/releases/detail/information-box.tsx
+++ b/application/frontend/src/pages/releases/detail/information-box.tsx
@@ -10,7 +10,7 @@ import { EagerErrorBoundary } from "../../../components/errors";
 type Props = {
   releaseKey: string;
   releaseData: ReleaseTypeLocal;
-  releaseDataIsFetching: boolean;
+  releaseDataIsLoading: boolean;
 };
 
 /**
@@ -19,13 +19,13 @@ type Props = {
  *
  * @param releaseKey the unique key referring to this release
  * @param releaseData the information about this release
- * @param releaseDataIsFetching
+ * @param releaseDataIsLoading
  * @constructor
  */
 export const InformationBox: React.FC<Props> = ({
   releaseData,
   releaseKey,
-  releaseDataIsFetching,
+  releaseDataIsLoading,
 }) => {
   const isAllowMutateActivation = releaseData.roleInRelease == "Administrator";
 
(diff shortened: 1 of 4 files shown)


DoxasticFox:
Not sure if this really does close https://github.com/elsa-data/elsa-data/issues/479. I didn't test this PR by running a job; I tested it by changing tabs, from Elsa to something else repeatedly. Doing that produced a visual artifact which looked the same as in the issue, which I fixed

andrewpatto:
So when a job is running - it starts a 5 (10?) second timeout event - which then forces that status refresh. Can you take a quick look and see if the placement of that is problematic in a mount/unmount sense (I suspect it is unmounting the entire tree - when it probably can just unmount that status bar itself)

But if there is nothing obvious - happy for this is be put through at least as a slightly better ui

DoxasticFox:
My understanding of the mechanics of the issue was that [this hook](https://github.com/elsa-data/elsa-data/blob/e285768a2aeada48aeb1d03f160217d2a119731a/application/frontend/src/pages/releases/detail/releases-detail-sub-page.tsx#L18) was causing everything in the `ReleasesDetailSubPage` component to be rendered whenever I'd toggle the tab. [I think that hook is also what's doing the 5-second refresh you're talking about](https://github.com/elsa-data/elsa-data/blob/e285768a2aeada48aeb1d03f160217d2a119731a/application/frontend/src/pages/releases/releases-master-page.tsx#L28).

In this PR, Elsa still triggers a re-render pretty frequently, but the renders don't cause the affected components to temporarily unmount because I replaced `releaseDataIsFetching` with `releaseDataIsLoading`. [TanStack Query's documentation](https://tanstack.com/query/v4/docs/react/reference/useQuery) says the difference between "loading" and "fetching" is that the former is like the latter, except "loading" only refers to the initial fetch.

application/frontend/src/pages/releases/detail/information-box.tsx
```diff
@@ -61,45 +61,12 @@ export const InformationBox: React.FC<Props> = ({
...
-          deactivateMutation.mutate(
-            { releaseKey },
-            { onSuccess: async () => await queryClient.invalidateQueries() }
-          )
-        }
-      >
-        Deactivate Release
-      </button>
-    </div>
-  );
```
DoxasticFox, line 95: Also, I inlined this function because otherwise React couldn't figure out that this was the same component with slightly different data between renders

williamputraintan:
I guess one way instead of invalidating all queries [here](https://github.com/elsa-data/elsa-data/blob/e285768a2aeada48aeb1d03f160217d2a119731a/application/frontend/src/pages/releases/releases-master-page.tsx#L71), we might just invalidate the running job only query, and then only invalidate all when running job is done (smth like below)?

```
  useEffect(() => {
    const interval = setInterval(async () => {
      if (releaseQuery?.data?.runningJob) {
        await utils.release.getSpecificRelease.invalidate();
      } else {
        await queryClient.invalidateQueries();
      }
    }, REFRESH_JOB_STATUS_MS);
    return () => {
      clearInterval(interval);
    };
  }, [releaseQuery?.data?.runningJob]);
```

mmalenic (approved):

PR #500, opened 2023-10-18
S3 Access Point Trackable Egress

### LINKED ISSUE #494: Get egress stats working with access point objects
With the newer access point sharing - there is no release identifier in the cloudtrail entry. However, there will be an access point bucket name that does map back to the particular release. Need to teach the egress routines to understand this.

Currently (understandably as access point sharing is new)

Screenshot 2023-10-10 at 2 20 28 pm

williamputraintan:
Fix #494 

- Bug fixes
- Fix CloudTrailLake query for AccessPoint via the ARN created
- Store AccessPoint ARN in `release::Activation` schema when a cloudformationInstallJob installed finished
- Add some logic when deactivating release
	- Cannot deactivate when running job exist
	- uninstalling access point when release deactivated
- Error when activating a release with no sharing mechanism
- Some UI addition

To consider:
- Select Access Point name should not update Db until it is installed

### DIFF
--- a/application/backend/src/business/services/jobs/job-cloud-formation-create-service.ts
+++ b/application/backend/src/business/services/jobs/job-cloud-formation-create-service.ts
@@ -254,8 +254,11 @@ export class JobCloudFormationCreateService extends JobService {
           await this.awsAccessPointService.getInstalledAccessPointObjectMap(
             cloudFormationInstallJob.forRelease.releaseKey,
           );
+
         const apArn = new Set<string>(
-          Object.values(map).map((ape) => ape.objectStoreBucket),
+          Object.values(map)
+            .map((d) => d.accessPointArn)
+            .filter((d) => !!d) as string[],
         );
 
         // Append with existing one if any
--- a/application/backend/src/business/services/jobs/job-cloud-formation-create-service.ts
+++ b/application/backend/src/business/services/jobs/job-cloud-formation-create-service.ts
@@ -89,7 +89,9 @@ export class JobCloudFormationCreateService extends JobService {
           status: e.job.JobStatus.running,
           started: e.datetime_current(),
           percentDone: e.int16(0),
-          messages: e.literal(e.array(e.str), ["Created"]),
+          messages: e.literal(e.array(e.str), [
+            "Triggered install of stack in AWS",
+          ]),
           auditEntry: e
             .select(e.audit.ReleaseAuditEvent, (ae) => ({
               filter: e.op(ae.id, "=", e.uuid(newAuditEventId)),
--- a/application/backend/src/business/services/jobs/job-cloud-formation-create-service.ts
+++ b/application/backend/src/business/services/jobs/job-cloud-formation-create-service.ts
@@ -36,7 +36,9 @@ export class JobCloudFormationCreateService extends JobService {
     @inject("CloudFormationClient")
     private readonly cfnClient: CloudFormationClient,
     @inject(AwsEnabledService)
-    private readonly awsEnabledService: AwsEnabledService
+    private readonly awsEnabledService: AwsEnabledService,
+    @inject(AwsAccessPointService)
+    private readonly awsAccessPointService: AwsAccessPointService,
   ) {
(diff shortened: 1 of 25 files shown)


application/backend/src/business/services/releases/release-activation-service.ts
```diff
@@ -124,9 +129,25 @@ export class ReleaseActivationService extends ReleaseBaseService {
         if (releaseInfo.activation)
           throw new ReleaseActivationStateError(releaseKey);
```
williamputraintan, line 131: I think it make sense to check if there is a sharing configuration enabled before starting a release?

application/backend/src/business/services/jobs/job-cloud-formation-create-service.ts
```diff
@@ -225,26 +229,67 @@ export class JobCloudFormationCreateService extends JobService {
...
-        this.edgeDbClient
+        this.edgeDbClient,
       );
 
       if (!cloudFormationInstallJob)
         throw new Error(
-          "Job id passed in was not a Cloud Formation Install Job"
+          "Job id passed in was not a Cloud Formation Install Job",
+        );
+
```
williamputraintan, line 248: So to store every AccessPoint's ARN, every cloudFormationInstallJob is done, it will query AcessPointArn and store it in `release::Activation`

application/backend/src/business/services/releases/release-activation-service.ts
```diff
@@ -197,15 +215,41 @@ export class ReleaseActivationService extends ReleaseBaseService {
         }
       },
       async (tx, _) => {
+        console.log("trigger tx");
         const { releaseInfo } = await getReleaseInfo(tx, releaseKey);
 
         if (!releaseInfo) throw new ReleaseDisappearedError(releaseKey);
```
williamputraintan, line 221: Just in case, I think we could disable deactivation release when there is an active job running.

application/backend/src/business/services/releases/release-activation-service.ts
```diff
@@ -197,15 +215,41 @@ export class ReleaseActivationService extends ReleaseBaseService {
...
 
+        // Not allowing to deactivate release while there is an existing running job
+        if (isRunningJob)
+          throw new ReleaseDeactivationRunningJobError(releaseKey);
+
         // importantly for this service we need to delay checking
         // the activation status until we get inside the transaction
         if (!releaseInfo.activation)
           throw new ReleaseDeactivationStateError(releaseKey);
```
williamputraintan, line 230: Also when deactivating release I think we should uninstall any active access point

williamputraintan:
Also adding some disabled text and inputs for existing running job
Screenshot 2023-10-19 at 17 16 39

application/backend/src/business/services/releases/release-activation-service.ts
andrewpatto: Yes - can you make an issue for doing that (agreed - was on my list of things to do)

williamputraintan: So I added a delete access point on deactivation right [under this comment](https://github.com/elsa-data/elsa-data/pull/500/files#r1364951210) (on[ line 231-250](https://github.com/elsa-data/elsa-data/blob/bafe3453c46254f2c86d1031310f2ed70c6a3b10/application/backend/src/business/services/releases/release-activation-service.ts#L231-L250)). But happy to create an issue for this if we need to refactor this out now.

andrewpatto (approved):

application/backend/src/business/services/releases/release-activation-service.ts
andrewpatto: Looks like a console log left here @williamputraintan

williamputraintan: Removed on the next commit ([bafe345](https://github.com/elsa-data/elsa-data/pull/500/commits/bafe3453c46254f2c86d1031310f2ed70c6a3b10)). Just that the comment was done before the commit

PR #512, opened 2023-11-03
Group Cases by patient proband Id

### LINKED ISSUE #509: Get the australian genomics data loader to handle orphaned family members
When some samples are promoted from singleton to trio - the original singleton does not have FAMxxxxx in its filename. This means that currently we don't put them into the same case grouping in Elsa Data.

However, for some datasets - all members of the family *do* have a consistent patient id in the form
Axxxxx
Axxxxx_mat
Axxxxx_pat

What we can do is after the initial scan for loading - we can detect situations where a family (of any family id) - has more than one patient in it with a consistent Axxxxx prefix. We can then reparent any loose patients into that family.

This should work to retrieve patients like
Axxxxxx

But might also be able to salvage entries like Axxxxx_sib that have also not correctly been named into a family

Some examples from the current cardiac load to prove the case:

FAM000524 has A1433005_pat and A1433005_mat - but there is a standalone A1433005 not in the family.

A1433014_sib exists by itself - but there is a FAM000422 it could belong to.

williamputraintan:
Fix #509

### DIFF
--- a/application/backend/src/business/services/australian-genomics/s3-index-import-service.ts
+++ b/application/backend/src/business/services/australian-genomics/s3-index-import-service.ts
@@ -728,6 +728,7 @@ export class S3IndexApplicationService {
           );
         const datasetConfig: any =
           this.datasetService.getDatasetConfiguration(datasetUri);
+
         if (datasetConfig) {
           if (datasetConfig.specimenIdentifierRegex) {
             datasetSpecimenIdRe = datasetConfig.specimenIdentifierRegex;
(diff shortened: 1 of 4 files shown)


williamputraintan:
Screenshot 2023-11-03 at 17 38 33

application/backend/src/business/services/australian-genomics/s3-index-import-service.ts
```diff
@@ -813,6 +814,37 @@ export class S3IndexApplicationService {
...
+      // the filename)
+      const probandIdMapCaseId: Record<string, string> = {};
+
+      // Need to loop through all studyId
+      for (const m of groupedManifestByFiletype) {
+        const patientIdArray = m.patientIdArray;
+        const filenameId = m.filenameId;
+
+        for (const patientId of patientIdArray) {
+          const probandId = patientId.split("_")[0];
```
andrewpatto, line 834: Is there any need we might have to make this configurable? I mean nothing too fancy - but what about if there is a flagship that doesn't use _mat and _pat..

i.e. for patient ids

AA_BB_1
AA_CC_2

Something where the underscores are incidental - but we might now accidentally attribute meaning to them.

If all the AG flagships are safe from this then that's fine for me (no need to add flag)

williamputraintan: I think for AG flagship study Id are safe from double underscores. As the validation pipeline studyId [regex](https://github.com/umccr/agha-data-validation-pipeline/blob/959e0285e42f2c7e6a73488de66f0545dc39149c/lambdas/layers/util/util/agha.py#L5) will only allow those values (`_mat`, `_pat`, ...)

andrewpatto (approved):

PR #516, opened 2023-11-07
Updates on Copy-Out Egress Tracking

williamputraintan:
- Found and fix a bug on manifest generation (which should be a seperate PR context)
- Added 100% egressing record after successful copyOut

### DIFF
--- a/application/backend/src/business/services/_release-file-list-helper.ts
+++ b/application/backend/src/business/services/_release-file-list-helper.ts
@@ -129,7 +129,7 @@ export const unpackFileArtifact = (
         });
       }
     }
-    console.log("rawArray", rawArray);
+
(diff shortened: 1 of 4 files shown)


application/backend/src/business/services/_release-file-list-helper.ts
```diff
@@ -40,96 +40,100 @@ export const unpackFileArtifact = (
   const getRaw = (): Pick<
```
williamputraintan, line 40: It seems to be a bug here, as it only select the first artifact file type appear but not the rest. In other words, index files were never returned when generating manifest.

DoxasticFox (approved):

andrewpatto (approved):

Issue #522, opened 2023-11-28
Plain manifest should have an OBJECTNAME column


andrewpatto:
We took the OBJECTURL out of the manifest as it revealed "too much" information. However, we do need the names of the files to do reconcilliation .. suggest putting in an OBJECTNAME column

PR #523, opened 2023-11-29
Added an object store name property that can appear in manifests

andrewpatto:
Added a manifest download for copy out


andrewpatto:
Basically for copy out - they end up with a bunch of files in a folder - but they need a way to match those file names to the case ids etc. They *don't* need the object URL (which has too much detail)

andrewpatto:
Closes #522

andrewpatto:
Rather than change the plain manifest - I actually added a manifest download for "Copy Out" with the extra column

williamputraintan (approved):

PR #524, opened 2023-12-18
Settings for gitpod

### LINKED ISSUE #478: Do all the bumping needed for Node20
I think is waiting on Node20 support at AWS level..

but then node 18 is mentioned in lots of places..

andrewpatto:
Setting a gitpod workspace
Along the way fixes #478 - because gitpod used Node 20 anyhow


andrewpatto:
I mean nothing much to review - the gitpod stuff is still work in progress (but does vaguelly work).
And the `tsx` stuff is just because ts-node and all the newer Nodejs seem to be broken together

.gitpod.Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+RUN echo deb [signed-by=/usr/local/share/keyrings/edgedb-keyring.gpg]\
+      https://packages.edgedb.com/apt \
+      $(grep "VERSION_CODENAME=" /etc/os-release | cut -d= -f2) main \
+      | sudo tee /etc/apt/sources.list.d/edgedb.list
+
+# Our list of tools we need to use for dev (that won't be installed via npm etc)
+# pre-commit is used for checking before git commits
+# edgedb is our database
+RUN sudo apt-get update && sudo apt-get install -y edgedb-3 pre-commit
```
DoxasticFox: [Docker's documentation](https://docs.docker.com/develop/develop-images/instructions/#apt-get) suggests the use of `rm -rf /var/lib/apt/lists/*` to reduce the container size

DoxasticFox (approved):

.gitpod.Dockerfile
andrewpatto: Yeah its an image inside the Gitpod infrastructure so I wasn't too worried about image size - but will add. Thanks!

mmalenic (approved):

.gitpod.Dockerfile
DoxasticFox: Ah, I figured you had a good reason. TIL

PR #525, opened 2023-12-19
Add a nightly docker image build

### LINKED ISSUE #341: Make our docker builds occur nightly (as well)
Would be good if the dev branch was published nightly as a docker image called

`nightly-YYYY-mm-dd`

would be useful for allowing our AWS deploys to be set to something vaguely up to date but immutable - without needing to force a release or futz about with the 'dev' branch that keeps changing

I think it is relatively trivial to add this to the github action (happy to abort if it isn't easy)

andrewpatto:


andrewpatto:
I haven't actually tested this - as I think schedules seem to only run off the main branch

DoxasticFox (approved):

andrewpatto:
This is not building due to the Node 20 changes (see other PR). So will have to be waved through..

andrewpatto:
Closes #341

PR #526, opened 2023-12-19
Can't remove apt cache

andrewpatto:
The gitpod base image actually has permissions that forbid this


DoxasticFox (approved):

andrewpatto:
actually - why don't I sudo!

PR #528, opened 2023-12-21
Applicant email is not required

### LINKED ISSUE #521: Allow manual releases to have no users
There are some manual releases that might be made completely by the AG staff (select files, copy out) - and that does not require the recipient of the data to be involved. We currently insist they add a user - but I suggest we can just take that out of the "create manual" dialog

andrewpatto:


andrewpatto:
Closes #521

